PR #3, opened 2018-03-03
Mention Post-CSS approach to adding new utilities

budparr:
I mentioned the error so it may surface in search.


budparr:
I don't this is as important now with improved [Post-css import support.](https://github.com/tailwindcss/tailwindcss/releases/tag/v0.5.0#better-postcss-import-support)

PR #17, opened 2018-04-18
Fixing classname extraction in templated strings

somebody32:
The full case described here: https://github.com/FullHuman/purgecss/issues/75

In result was not a purgecss issue but extractor being too eager and matching class names together with `.


somebody32:
closed in favor of https://github.com/tailwindcss/docs/pull/18

PR #49, opened 2018-10-13
Suggest using `npx` command instead of local bin

qrazi:


adamwathan:
It ships with npm these days so I’m tempted to merge but won’t work for
people who use Yarn I don’t think? 🤔

m1guelpf:
I mean, you could also `yarn tailwind`, but that will not work for people without yarn. I'd say leave the one that works for everyone (current)

qrazi:
Reading all the comments I think I might have gotten carried away by my discovery of the `npx`-command a bit too much...😅

I did assume that anyone using this project would be using `node` & therefor `npm` and therefor practically everyone would also have `npx`. But I now realize those are a bit too brought sweeping assumptions. I guess this PR could be recovered by rephrasing to something similar as the [installation](https://tailwindcss.com/docs/installation#1-install-tailwind-via-npm) itself, which mentions both `npm` and `yarn`. Or decline this PR, since the instructions weren't broken in the first place?

adamwathan:
Merged an earlier PR suggesting the same change, will add a second line with instructions for yarn users.

PR #74, opened 2019-02-07
Use correct package name for npx commands

ynnoj:


adamwathan:
This doesn't work, you have to type `npx tailwind` not `npx tailwindcss` because the binary is called `tailwind`:

image

ynnoj:
Weird, I see the opposite.

screenshot 2019-02-07 at 16 29 15

PR #94, opened 2019-03-19
Added custom font class back to the body tag

jalendport:
I'm assuming we still want Source Sans Pro as the font for the docs? The `source-sans` class has been migrated to the new Tailwind config at least...


jalendport:
Gotcha 👍

PR #179, opened 2019-05-04
Use rgba over color keywords, for macOS Safari

assertchris:
macOS Safari seems to prefer Reba over the keywords in linear gradient


assertchris:
Fixed with https://github.com/tailwindcss/docs/commit/879a94dfcd1d760b68cbd6f00cadcbc182b2daaf

PR #237, opened 2019-06-11
Document config utilities

AlexVipond:
Stumbled across the `negative` function while digging through the default theme—thought it would be good to document it and make a nice space in the docs for any config utilities that get exposed in the future.


CvX:
(Just a heads-up: there’s an accidental lock file in the PR 😃)

AlexVipond:
@CvX I deleted it, good catch!

PR #242, opened 2019-06-14
fixed commands to run tailwind via npx

MZanggl:
Changing from `npx tailwind` to `npx tailwindcss`.


MZanggl:
I get it, you first need to install tailwindcss through npm, then `npx tailwind` works just fine.

PR #246, opened 2019-06-20
Fix multiple active sidebar items

### LINKED ISSUE #245: Multiple active items in sidebar content listing
I noticed when I went to the [Flex Wrap section of the docs](https://tailwindcss.com/docs/flex-wrap), that both the Flex Wrap and Flex items are visually highlighted in the sidebar content listing. They are not adjacent items, so it's not because they're both in view. It seems to happen on docs for anything that starts with 'Flex' (e.g. 'Flex Direction', 'Flex Grow'). Other items could be affected. 

Screenshot of the Flex Wrap documentation where Flex Wrap and Flex items are highlighted in the content list.

Maybe it's because the check for if a page is active is using `starts_with`?
https://github.com/tailwindcss/docs/blob/0dfd9645b2f4c0c4e7c643a6aa6cf0c4a4be4156/config.php#L30

Also, I'm not sure if it's necessary to make an array for a page and loop through it in that function.
Seems unnecessary, but maybe `collect()` does something more than it appears.

missmatsuko:
Also removes creating an array from and looping through $page in the same function.

Closes #245

### DIFF
--- a/config.php
+++ b/config.php
@@ -10,7 +10,7 @@ return [
     'config' => json_decode(file_get_contents(__DIR__ . '/tailwind.json'), true),
     'version' => json_decode(file_get_contents(__DIR__ . '/node_modules/tailwindcss/package.json'), true)['version'],
     'colors' => ['red', 'orange', 'yellow', 'green', 'teal', 'blue', 'indigo', 'purple', 'pink'],
-    'activeCategory' => function ($page) {
+    'categoryName' => function ($page) {
         if (starts_with($page->getPath(), '/docs')) {
             return 'Documentation';
         } elseif (starts_with($page->getPath(), '/components')) {
--- a/config.php
+++ b/config.php
@@ -26,6 +26,11 @@ return [
 
         return str_is($page->getPath(), $path);
     },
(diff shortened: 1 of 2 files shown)


adamwathan:
Hey thanks for this! The only issue I think is that it will break the active styles on the parent nav items, like you can see here that "Documentation" is bold whenever any of the documentation pages are active:

image

Probably we just need to switch to a different helper function for the category highlighting altogether, like `categoryActive` and have that check the start of the string instead of the whole string, and use exact matching for the regular sidebar items?

missmatsuko:
Thanks for pointing that out. I've fixed that functionality now.

I did notice that the sidebar items don't get highlighted when they're in-page links, e.g. https://tailwindcss.com/resources#learning-design

Probably would be good to highlight those as well, but I think it might require JavaScript? I'm not sure if it can be done just in the template.

adamwathan:
This is great, thank you @missmatsuko! 🙌🏻

Yeah getting the sidebar highlighting working for in-page links would be a bit of trickery with JS, not a huge priority I don't think but of course would welcome a PR if anyone wants to try and tackle it.

PR #255, opened 2019-07-14
Add note about Chrome placeholder bug to position docs

GeoffSelby:
This adds a note about Chrome's placeholder bug pertaining to absolute and fixed positioned elements with a recommended quick fix.

There is a bug in Chrome that places placeholders above all other elements as if it has a higher z-index. This causes them to sit on top of fixed elements when scrolling past. A quick fix for that is to explicitly set the z-index of fixed elements to anything over 0. With Tailwind, we can use `.z-10` to achieve this. It's hacky, but worth noting.

This PR adds a note to the `Absolute` and `Fixed` section of the docs explaining the hack.

### DIFF
--- a/source/docs/position.blade.md
+++ b/source/docs/position.blade.md
@@ -95,7 +95,26 @@ Offsets are calculated relative to the element's normal position and the element
 
 Use `.absolute` to position an element *outside* of the normal flow of the document, causing neighboring elements to act as if the element doesn't exist.
 
-Offsets are calculated relative to the nearest parent that has a position other than `static`, and the element *will* act as a position reference for other absolutely positioned children.
+<div class="text-sm bg-blue-100 text-blue-700 font-semi-bold px-4 py-2 mb-4 rounded">
+  <div class="flex items-center">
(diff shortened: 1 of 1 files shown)


DCzajkowski:
Hey, I don't quite understand what you mean and it is not clear enough in mu opinion in the docs. I am also not sure if Tailwind documentation is a good place to describe bugs in specific browsers. I think it's better to report this in bugs.chromium.org rather than documenting in a CSS framework docs. 😉

GeoffSelby:
The bug was previously submitted to Chromium. I guess I could have done a better job explaining what the cause is. In Tailwinds preflight after including normalize.css, there is a block of code for normalizing placeholder styles. That block sets the opacity of the placeholder and for some reason causes the bug. For example, if you have a fixed navbar that sticks to the top of your page, and a form with placeholders in the main section... the placeholder itself would sit above the navbar if scrolling past. It only effects Chrome/Chromium though and is admittedly very rare to come across so after some thought I’m closing this PR.

PR #259, opened 2019-07-22
Refine purgecss defaultExtractor regex for vuejs

oliverpool:
A Vue.js component can have the following attribute: `:class="{ invisible: someCondition }"` (sometimes even enforce by some linters)

With the current suggested regex `[A-Za-z0-9-_:/]+`, the whole `invisible:` word is interpreted as a class and the `invisible` class will be stripped away.
Changing the regex to prevent the `:` from being at the last position solves this issue: `[A-Za-z0-9-_:/]*[A-Za-z0-9-_/]`

I am not aware of any tailwind class ending with `:` (which this MR would break) and I think that someone generating such classes intentionally should be skilled enough to also adapt this regex to its usecase :-)

### DIFF
--- a/source/docs/controlling-file-size.blade.md
+++ b/source/docs/controlling-file-size.blade.md
@@ -62,7 +62,7 @@ const purgecss = require('@fullhuman/postcss-purgecss')({
   ],
 
   // Include any special characters you're using in this regular expression
-  defaultExtractor: content => content.match(/[A-Za-z0-9-_:/]+/g) || []
+  defaultExtractor: content => content.match(/[A-Za-z0-9-_:/]*[A-Za-z0-9-_/]/g) || []
 })
 
 module.exports = {
--- a/source/docs/controlling-file-size.blade.md
+++ b/source/docs/controlling-file-size.blade.md
@@ -92,7 +92,7 @@ const purgecss = require('@fullhuman/postcss-purgecss')({
 The way it works is intentionally very "dumb". It doesn't try to parse your HTML and look for class attributes or dynamically execute your JavaScript — it simply looks for any strings in the entire file that match this regular expression:
 
 ```js
-/[A-Za-z0-9-_:/]+/g
+/[A-Za-z0-9-_:/]*[A-Za-z0-9-_/]/g
 ```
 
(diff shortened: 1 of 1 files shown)


adamwathan:
Ah nice I think this is a good addition 👍 I wonder if there's any way to avoid the duplication in the regex, feels like must be some way to simplify it but my pattern-foo isn't quite there personally, haha.

oliverpool:
I see two possibilities to "simplify" the regex:
1. replace `A-Za-z0-9_` with `\w`: `[\w-:/]*[\w-/]`
2. only match classes which ends with an alphanumeric character `[\w-:/]*\w` (I don't think that Tailwind generates classes ending with `-` or `/` (or `:`)

So the compacter expression that I can think of would be `[\w-:/]*\w` (but I am not sure if it is readable enough. Alternative: `[A-Za-z0-9-_:/]*[A-Za-z0-9]`)

What do you think?

adamwathan:
This seems to work too, using a negative lookbehind assertion:

```
[A-Za-z0-9-_:/]+(?<!:)
```

Not sure if it's bad for performance or anything though 🤔

oliverpool:
> my pattern-foo isn't quite there personally

Well, it is much better than mine (I had never heard of "negative lookbehind assertion")

I compared the regex and they seem to behave the same.

I also run some tests on https://jsbench.me/ and I got following results:

image

So it seems that your regex is about 5-10% faster :tada:

Should I replace the regex in my MR with this new regex ?

```
[A-Za-z0-9-_:/]+(?<!:)
```

tlgreg:
Just a note, for a lookbehind you do need node `9.11.2+`

adamwathan:
Hm that sucks about the node version since node 8 is still pretty prevalent. Not sure on the best thing to do here.

oliverpool:
Then we can recommend `[\w-:/]*\w` in the meantime.

I think that fixing the initial issue (`:class="{ invisible: someCondition }"` wrong class extraction) is quite important for the doc. The exact regex is not that important.

IMHO most of the people will only copy and paste it, hoping that it works. People who wants to understand it and customize it, are also able to fix eventual bugs.

adamwathan:
I'm good with recommending `[\w-:/]*\w` if you want to update the PR with that! Thanks 👍

adamwathan:
Made the tweaks myself, thanks again!

PR #298, opened 2019-10-13
Plugins on configuration page need extra brackets

BreezeZin:
I have tested all of these plugins and they all need the brackets to work.


tlgreg:
It depends on the plugin.

PR #299, opened 2019-10-14
Note the browser caching benefits of using Tailwind over inline styles

imliam:
I think that browser caching, while less important than the other points mentioned are to maintainability of your source code, is an important distinction to make between Tailwind and inline styles.


simonswiss:
Hey @imliam 👋

This is definitely a valid point - but it doesn't really fit in the current context of the page, within the "why not inline styles" and above the responsive code example.

The code example demonstrates the 3 other points: constraints, responsive, pseudo selectors.

Definitely not as visible, but Nicholas Gallagher's article that talks about Gzipping perf is linked from almost all the resources shared in the bottom of the page.

Going to close this PR, but we'll keep in mind that there is an opportunity to talk about browser caching (on the HTML side particularly) somewhere in the docs.

Thank you!

PR #303, opened 2019-11-02
Fix npx command

remotelydev:
npx tailwindcss init works 
npx tailwind init does not


adamwathan:
`npx tailwind init` does work as long as you've installed Tailwind first:

image

adamwathan:
Ended up updating this after we aliased the command to `tailwindcss` so that both would work. Thanks for the suggestion sorry for not merging this and doing it myself 🤦

PR #325, opened 2019-12-24
Add note for loading Inter font

GeoffSelby:
This PR adds a note about loading `Inter` from the cdn as it will not be included in Tailwind V2 per [#1275](https://github.com/tailwindcss/tailwindcss/pull/1275).

This PR does not load `Inter` as it will not be added to the `sans` font stack until V2. Once V2 is ready to be released, we can add the cdn link to load the font for the example.


bradlc:
Hey @GeoffSelby. Inter is no longer a part of the default font-family config so I don't think this note is necessary. Thanks!

PR #326, opened 2019-12-26
Add autofocus to search input field

karllson:
This PR adds autofocus to the search bar on the main page of tailwindcss (not on subpages). 

I think there might be a high interest in searching the docs when visiting the main page. In addition to that, this change won't bother other users who are not interested in exploring the docs.

The search fields of the subpage (Documentation, Components, ...) won't be autofocused because the user interests on these pages should be slightly different.


bradlc:
Hey @karllson, thanks for the PR! I'm going to close this as the search is not the primary purpose of the page, so autofocusing the search input doesn't seem appropriate.

PR #334, opened 2020-01-18
fixed typo

jtuller:


jtuller:
/facepalm... this is not a type - it is NEGATIVE MARGIN!!! My mind is blown. going to go back to work now...

PR #357, opened 2020-02-11
Fix config import

aparajita:
When using webpack/babel, to import the default export of a commonjs module in ES6, you must use the syntax `import * as foo from 'module.js'`.

### DIFF
--- a/source/docs/configuration.blade.md
+++ b/source/docs/configuration.blade.md
@@ -476,7 +476,7 @@ To make this easy, Tailwind provides a `resolveConfig` helper you can use to gen
 
 ```js
 import resolveConfig from 'tailwindcss/resolveConfig'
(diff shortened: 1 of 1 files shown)


aparajita:
Hmm... this works at compile time, but is breaking at runtime. Need to find a solution.

aparajita:
I've searched all over, and can't find a solution that works. If I used the proposed syntax, it compiles, but at runtime I'm getting the error:

```
tailwind.config.js:16
Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'
```

Any suggestions?

aparajita:
Figured it out. I'm using nuxt, so this had to go in `nuxt.config.js`:

```js
  build: {
    babel: {
      sourceType: 'unambiguous'
    }
  }
```

PR #386, opened 2020-03-19
Update CDN link

martbock:
If there is a specific reason to reference an older version of tailwindcss on the CDN, please disregard this PR.

Otherwise, this pull request just bumps the version number in the CDN link up to `1.2`.


martbock:
I have just noticed that the URL updates automatically and does indeed represent the current version. 🙈

PR #392, opened 2020-04-02
Ordering Default color palette to mirror Tailwind UI

tjkohli:
This will be helpful for people who may want to compare the palettes or find overrides. I wasted some time figuring out why my overrides weren't working correctly between TailwindCSS and Tailwind UI — when I realized both documentations did not present the colors in the same order!


simonswiss:
Hey - thanks for suggesting this!

I looked into getting this PR merged, but since the rebuild to NextJS the code for colors has changed from `blade` to `JSX`, and the colors are now matching the Tailwind UI color order already.

Won't be able to merge it sorry 😢 - but thank you for your contribution!

PR #439, opened 2020-06-02
Add note for default key when using theme function

JohBa:
I was a bit confused that the `default` key is special for class generation but not for theme function so I think a small note here could be useful


simonswiss:
Thanks for this 👍

Updated your fork to new NextJS site, reworded your copy a little, and going to merge it now! 🎉

PR #450, opened 2020-06-15
Update plugins.blade.md

mmavko:
Fixing code in a comment.

Fixing escaping in an example.


adamwathan:
The escaping example is actually correct as weird as that seems! Will tweak that back and merge though thanks.

image

mmavko:
Ha-ha, that's really confusing!

Probably it makes sense to add a comment like they do it on [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape):

```js
CSS.escape(0) // "\30 ", the Unicode code point of '0' is 30
```

PR #463, opened 2020-07-09
Fix typo in border-color.blade.md

pdziewa:


pdziewa:
Sorry! Just found #443

PR #478, opened 2020-07-23
Fix: Do not use bg where border is meant

bodograumann:
This was probably introduced by a copy and paste.


bodograumann:
Ugh. Duplicate of #443. Please merge that one!

PR #485, opened 2020-07-30
Fix Blog navlink showing active state when on Resources page

asmartdev:
Current behaviour gives the Blog link in the side navigation an active state when on the Resources page. This fixes the issue by removing the active state check - the blog is a separate site so there isn't a page which would need to show the Blog page as active.


bradlc:
Thanks @asmartdev! This ended up being fixed during the Next.js rebuild of the site

PR #504, opened 2020-08-25
Adds page for transform utility

nerdcave:
The doc page for `transform` has never existed, as far as I can tell. I listed it first under the `Transforms` section since other transform utilities are dependent on it. 🤷‍♂️

I didn't add any specific responsive examples since that seemed like overkill for this.

If a doc page for `transform` was omitted intentionally, feel free to just close this out.

### DIFF
--- a/src/layouts/DocumentationLayout.js
+++ b/src/layouts/DocumentationLayout.js
@@ -137,6 +137,7 @@ const nav = {
     pages['animation'],
   ],
   Transforms: [
+    pages['transform'],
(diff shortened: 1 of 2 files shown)


simonswiss:
Hey, thanks for the suggestion!

Following an internal discussion, we've actually opted to not add a page dedicated to `transform`. We're going through a documentation website refresh in preparation for Tailwind 2.0, and we'll try be a bit more clear on that front in the new site 👍

PR #509, opened 2020-08-26
Replace unit of measurement K with KiB (kibibyte)

HonkingGoose:
## Changes:

- Replace user facing unit of measurement K with KiB (kibibyte)

## Context:

I think we should use kibibytes here, as we're talking about multiples of 1024 bits, we even do a conversion in the utility that is explicit about the unit used: ``const convertToKB = (bytes) => (bytes / 1024).toFixed(1) + 'K'``

Using the standard kibibyte unit removes the the ambiguity of the K unit. Because is K = 1000 bytes, or is K = 1024 bytes?
Using kilobytes instead of kibibytes is still ambiguous, as a kilobyte has historically been 1000 bytes or 1024 bytes.

[Wikipedia page on the Kibibyte](https://en.wikipedia.org/wiki/Kibibyte)

### DIFF
--- a/src/utils/stats.js
+++ b/src/utils/stats.js
@@ -9,7 +9,7 @@ const brotliSize = require('brotli-size')
 const CleanCSS = require('clean-css')
 const { loopWhile } = require('deasync')
 
-const convertToKB = (bytes) => (bytes / 1024).toFixed(1) + 'K'
+const convertToKB = (bytes) => (bytes / 1024).toFixed(1) + 'KiB'
 
 module.exports = () => {
   const stats = {}


adamwathan:
Appreciate the suggestion! I think I would personally prefer to just update things to accurate kilobyte numbers, since I have never heard of a kibibyte (cool to learn about though!) and Chrome dev tools reports asset sizes in kilobytes, so I think that's the unit people expect these numbers to be in.

HonkingGoose:
@adamwathan I'm a bit confused by this:
>  I would personally prefer to just update things to accurate kilobyte numbers,

Is there some problem with the current numbers? Or what do you mean with this?

adamwathan:
Hey sorry! I mean instead of updating to KiB, I think we should just divide by 1000 and use `kB`, so we are using the same units people are used to seeing in their browser dev tools:

image

Happy to merge that change if you'd like to make it 👍🏻

HonkingGoose:
I'll go make the wanted change in a new pull request then. :smile:

HonkingGoose:
Closing issue, superseded by #510.

PR #534, opened 2020-09-16
Typo fix

stevenschmatz:
Thanks for building and maintaining this great framework! Just noticed this as I was reading through the docs. These were the only two instances of the ".." I could find.


simonswiss:
Thanks so much for spotting and fixing this! 👍💯

PR #538, opened 2020-09-26
Make minor grammar tweaks to docs

nick-f:
Minor grammar tweaks to overflow, place-self, and space docs.


simonswiss:
Good stuff, thanks for these improvements! 👍

✅

PR #546, opened 2020-09-30
Fix Z-Index Required Code

### LINKED ISSUE #544: Z-Index page is missing info

IzMichael:
Resolves #544


simonswiss:
Hey!

See comment in #544 - don't think we should add the `absolute` classes in the code snippets here. I've explained why in the issue.

PR #551, opened 2020-10-01
Update upgrading-to-v1.mdx

MrSunshyne:
Move the following line to appear before the table with the label changes. 
> **These names need to change in the `variants` section as well, so feel free to do a find and replace across the whole file**.

**Why ?**
People will most likely be reading and executing this guide step by step. 
Someone might replace all the labels under "theme" only to realize after reading the next line that they have to do it again under "variants" (true story)


simonswiss:
Agree, and can see how this would lead people to save some time knowing to update both 👍

PR #552, opened 2020-10-01
Update tailwind.config.js appearances

MichaelDeBoey:
The `future` key was missing


simonswiss:
Yeah, good point, this has been added recently.

Wondering if we should also have the commented out flags, just like the output of the default Tailwind config file is... but it probably adds more noise than necessary - so I'll keep it like that 👍

PR #554, opened 2020-10-02
Migrate away from deprecated way of adding plugins to PostCSS

joiglifberg:


bradlc:
Hey @joiglifberg. I can't find any information about the `require` method being deprecated. In fact the `postcss` repo uses that method in its examples: https://github.com/postcss/postcss#usage

Additionally, it seems that the proposed change (using strings inside of a `plugins` array) is not valid: https://github.com/postcss/postcss-load-config#array

> ⚠️ When using an `{Array}`, make sure to `require()` each plugin

Going to close this but happy to take another look if I've missed something 👍 Thanks!

PR #559, opened 2020-10-10
fix html for horizontal card component example

n1ckdm:
the  element was missing a closing "/"


bradlc:
Hey @n1ckdm. The example is correct as it is since `img` is a [void element](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), although your version will also work fine 👍

n1ckdm:
thanks @bradlc - I think from memory I was trying to use the example in either JSX or a vue.js template and it wasn't working for me. It was fairly straightfoward to figure out I guess. Might be worth testing though incase anyone else has a similar probelm, I'll try putting a small example together and create an issue.

PR #573, opened 2020-10-22
add linear easing example

browner12:


simonswiss:
Hey, thanks for this!

The only problem with it is it makes the example a little "busy" with 4 buttons, and it completely breaks on mobile.

image

I thought about putting the buttons on two rows on mobile, or make the text shorter, but... ultimately anything I could come up with felt like it was _damaging_ the example more than the value added by the presence of the `linear` example.

Going to close this PR, but thank you so much for the suggestion 🙏 

Agreed it feels weird when we have all but one variants in the example, but... folks can still see the complete list of options in the table at the top of the page 👍

PR #579, opened 2020-10-29
Some missing single quotes in configuring-variants.mdx

david-treblig:
They were bugging me when I was reading the doc.


simonswiss:
Thanks for picking that up! They are bugging me too, now that I see them 😅

🚀

PR #586, opened 2020-11-18
Fix broken link to section

skjelmo:
The section titled «Using Tailwind without PostCSS» has `post-css` and not `postcss` in the [working link](https://tailwindcss.com/docs/installation#using-tailwind-without-post-css) to section `#using-tailwind-without-post-css`.


bradlc:
Thanks!

PR #591, opened 2020-11-19
Compress couple images

pavelloz:
It should help with performance a little bit.

image


pavelloz:
I also think some of the images could use .webp version of them just like couple images already have webp variants, because there is couple 100+KB jpegs that are good candidates for a shrinkage :)
 
image

adamwathan:
Thank you! 🙌

eneadriancatalin:
@pavelloz May I know what tool you used to compress these images? Thanks

pavelloz:
@eneadriancatalin the tool from first post is [ImageOptim](https://imageoptim.com)
And i think i created webp versions using [webp CLI tool from homebrew](https://formulae.brew.sh/formula/webp). Nowadays i would probably use AVIF instead of webp though :)

PR #592, opened 2020-11-19
Fix broken link

mattferderer:
Camel Case in the headers are changed to kebab case anchor links. This link needs the extra dash to work.


bradlc:
Thanks!

PR #594, opened 2020-11-19
Use correct casing for boxShadows default key

mikkelhl:
Changed the key for default boxShadow to be uppercased in the shadow example so it uses the correct key from the default config.

**Before**
before

**After**
after


bradlc:
Ah nice, missed that when upgrading to v2, thank you!

PR #599, opened 2020-11-19
Updates the default theme config url

Naoray:


bradlc:
Thanks @Naoray! Please could you update this to use `master` instead of `v2.0.0` so that it's the latest version?

Naoray:
@bradlc done

bradlc:
Thank you!

PR #601, opened 2020-11-19
Adding Angular Example to Component Driven Section

jneterer:
This PR is adding an example of Angular to the component driven section on the home page.

Tailwind is **_really_** growing in the Angular community! We have some awesome open source projects that make integrating Tailwind quick:

* https://www.npmjs.com/package/@ngneat/tailwind
* https://www.npmjs.com/package/ng-tailwindcss

and how-to articles are performing super well:

* https://dev.to/seankerwin/angular-8-tailwind-css-guide-3m45
* https://medium.com/@jacobneterer/angular-and-tailwindcss-2388fb6e0bab

so I am hoping this addition will show more Angular folks that Tailwind is not just for React and other frameworks, but works super well with Angular too!

Thanks for all your work and making such an amazing product. If approved, let me know what revisions I need to make!

Screen Shot 2020-11-19 at 8 05 30 AM

### DIFF
--- a/src/components/home/ComponentDriven.js
+++ b/src/components/home/ComponentDriven.js
@@ -9,6 +9,7 @@ import { ReactComponent as Icon } from '@/img/icons/home/component-driven.svg'
 import { Tabs } from '@/components/Tabs'
 import { ReactComponent as ReactLogo } from '@/img/icons/react.svg'
 import { ReactComponent as VueLogo } from '@/img/icons/vue.svg'
+import { ReactComponent as AngularLogo } from '@/img/icons/angular.svg';
 import { ReactComponent as LaravelLogo } from '@/img/icons/laravel.svg'
 import { AnimatePresence, AnimateSharedLayout, motion, useIsPresent } from 'framer-motion'
 import clsx from 'clsx'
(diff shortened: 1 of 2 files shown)


src/components/home/ComponentDriven.js
```diff
@@ -144,6 +145,134 @@ export default function Recipes({ recipes }) {
...
+export class NavComponent {}`).lines,
+    'nav-item.component.ts': tokenizeWithLines.jsx(`import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'app-nav-item',
+  template: \`
+    <li>
+      <a 
+        [routerLink]="link"
+        class="block px-4 py-2 rounded-md {{isActive ? 'bg-amber-100 text-amber-700' : ''}}"
```
ahasall, line 199: What about using `NgClass` here instead of the ternary ?

jneterer: @amirhosseinab You certainly could, but since this is just a simple example for the home page I am doing it this way. If you do it using ngClass while having `enablePurge: true` you would have to whitelist those classes. There's no way to show/explain this in this example, so I'm doing it this way for simplicity.

ahasall: I'm in favor of simplicity, but the following code should work with Purge CSS enabled:
```
<a [ngClass]="{ 'bg-amber-100 text-amber-700': isActive }" class="block px-4 py-2 rounded-md"></a>
```

jneterer: @ahasall It seems you're right, I didn't think the regex would catch the class inside an ngClass, but it looks like it does. I'll update it now! Thanks!!

jneterer:
Hello! Just checking in, curious to know if/when this might be merged and deployed? Thanks!

jneterer:
@simonswiss Hello! My PR has been open the middle of November so just wanted to check and see if/when this might be merged and deployed? Anxious to hear! Thanks for your time!

adamwathan:
Hey thanks for this! Suggestions like this that impact highly designed pages like the main marketing site can be a bit slow to merge because there are design considerations we need to make, and it usually involves multiple people on the team making time to look at something like this.

A few quick examples of what I mean:

- The Angular logo icon is a different color than all of the other framework icons, so we wouldn't want to merge this as is, but Steve likely will want to play with this himself to get the color right since the surface area of the icon is quite large so it will be perceptually darker than the other icons even if we use the same color
- The whitespace between the Blade icon and the code editor is quite tight, we probably need to rework all of the spacing to land on something we are happy with
- Having 4 icons becomes very tight on mobile, we probably want to re-think how this works as well

We are a very design-driven company so details like this are very important, especially on our landing page. There's a chance we may just have to stick with 3 examples due to the issues that arise with 4 unless we can dedicate the resources to a larger scale redesign of this section.

We will try to look at it and make a decision about it soon ❤️

jneterer:
Thanks @adamwathan for the response! I totally understand, that makes complete sense! Looking forward to a decision! And thanks for such a great product!!

adamwathan:
Hey folks! In case you didn't see we got this incorporated for the new redesign:

image

We messed up the commit history when merging things back in from our private repo so unfortunately the contributions aren't captured — really, really sorry 😞 Thanks again for this though.

jneterer:
@adamwathan Thanks so much for taking the time to follow up on this! No problem on commit history and contribution, the most important thing here is that Angular is featured, and I'm super grateful for that! 🥳  Many thanks to you and your team for such a great product!

PR #603, opened 2020-11-19
Remove duplicated disabled attribute

jofrly:
A `button` in the docs demonstrating the disabled state has the `disabled` attribute twice.


bradlc:
Thanks!

PR #607, opened 2020-11-19
Add webp variant of images

pavelloz:
Per https://github.com/tailwindlabs/tailwindcss.com/pull/591#issuecomment-730070146

Couple KB can be saved, but I dont want to make any decisions for you by switching to webp (just in case you want to support legacy) and I dont know next.js enough to make it with a fallback.

Maybe you will use them, if not, just close PR :)

Results are promising:
image

PS. `src/img/workcation-screenshot.jpg` is not used anywhere in this project (i didnt remove it because maybe its hotlinked somewhere outside)


bradlc:
Thanks @pavelloz, it's great to see the savings we could get with webp! I think the ideal solution would be to use the [Next.js Image component](https://nextjs.org/docs/api-reference/next/image), that way we don't need to maintain the different versions manually, or slow the build process down with conversions at build-time.

pavelloz:
> or slow the build process down with conversions at build-time.

I agree, thats why i generated them offline and commited to repo.

PR #612, opened 2020-11-20
Fix dark mode snippet

velut:
Correctly check that 'theme' is not present in localStorage.

I think it could be further simplified by saying `!localStorage.theme` as in 
```js
if (localStorage.theme === 'dark' || (!localStorage.theme && window.matchMedia('(prefers-color-scheme: dark)').matches)) {
    ...
}
```


bradlc:
Thanks @velut! I'm going to close this as it looks like this was already fixed 👍

PR #621, opened 2020-11-22
Fix PostCSS configuration example

tkiryu:
Fixes the typo for `postcss.config.js` in [Using a different file name](https://tailwindcss.com/docs/configuration#using-a-different-file-name) section.


bradlc:
Thanks!

PR #625, opened 2020-11-24
Use green instead of (non default) teal colour in text gradient example

vincentdoerig:
Spend around 20 minutes figuring out why the snippet [here](https://tailwindcss.com/docs/background-clip#cropping-to-text) didn't work on my project until I realised that `teal` isn't a default colour anymore. IMO it makes more sense to use the default colours in the examples to avoid such confusion.


simonswiss:
This is a good catch and this is actually what we were *meant* to do 😅

Thanks for doing this change 🎉

PR #636, opened 2020-11-29
Separate Tailwind directives from custom CSS in Next.js example

pacocoursey:
This is an important step for speeding up Next.js recompiling at development time.

If you write your custom CSS in the same `.css` file as the directives (particularly `@tailwind utilities`) the recompiles can take up to 10-15 seconds when you modify CSS. Separating into different files drops this back down to near-instant recompiles.

I updated the guide to recommend this happy path.


pacocoursey:
Closing re [this Twitter thread](https://twitter.com/adamwathan/status/1333160716178747401?s=20)

PR #643, opened 2020-11-30
Fix horizontal scrolling on the overflow examples

27pchrisl:
Adds inline-block to the paragraph, so the right-hand-side is correctly padded


27pchrisl:
Visible here: https://tailwindcss-com-git-overflow-horizontal-scroll-padding.tailwindlabs.vercel.app/docs/overflow#scroll-horizontally-if-needed

simonswiss:
Nice one, thank you! 👍

PR #644, opened 2020-11-30
Update overflow scroll examples to better highlight the expected behaviour

27pchrisl:
In the examples:
https://tailwindcss.com/docs/overflow#scroll-horizontally-always
and
https://tailwindcss.com/docs/overflow#scroll-vertically-always

They don't really show the difference between "if needed" and "always" because the text is so tall or wide as to make scroll bars always needed. This change shortens the text so it's clearer in the always case that the scroll bars show up when not needed.


simonswiss:
Nice improvement - that indeed conveys the message better, thank you 👍

PR #645, opened 2020-11-30
Modify scroll margin snap rule to better support future Safari

27pchrisl:
Currently the docs scroll to fragment targets such as https://tailwindcss.com/docs/overflow#visible using the `scroll-margin-block` property to avoid the fixed top search bar. This is not supported in Safari.

The `scroll-margin-top` property is supported in Safari TP (and other browsers), so will likely be in the next release of Safari (https://bugs.webkit.org/show_bug.cgi?id=189265#c12).

This change switches to using `scroll-margin-top` to support future Safari and continue supporting Chrome/Edge/Firefox.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -66,7 +66,7 @@ module.exports = {
               letterSpacing: '-0.025em',
             },
             'h2, h3': {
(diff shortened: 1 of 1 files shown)


bradlc:
Thank you!

PR #650, opened 2020-12-01
Add the regex that can be used to find and replace all leading dots

hivokas:
This regex would be useful for people who have a huge amount of leading dots in the project.


simonswiss:
Hello!

Thanks for that - that can indeed help a few folks 👍

I have reworded the sentence a little bit and going to merge this!

hivokas:
> I have reworded the sentence a little bit and going to merge this!

Thanks!

PR #653, opened 2020-12-02
Update gatsby.mdx

brainrepo:
Add the gatsby-config postcss missing step


simonswiss:
Good stuff, thanks for that! 👍

PR #659, opened 2020-12-07
Add border variant to checked example

alexanderbluhm:
In the example for the checkbox background- and border color are used.
However, only the background color variant is extended. So if you copy the example checkbox input and the config, the transparent border won't work.
By adding the border variant the `checked:border-transparent` will work.
There was also a bit confusion in the Discord, that's why I created this pull request.


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

PR #662, opened 2020-12-07
Fix: Search Magnifier Icon.

melbahja:
Search icon repeated
image


bradlc:
Hey thanks @melbahja. I think a better fix for this would be to add `background-size: 100%` instead, so that the icon is the correct size and aligned correctly.

Out of curiosity, which operating system and browser you are using?

melbahja:
Chrome on 1920x1080 screen.

bradlc:
Thank you!

PR #664, opened 2020-12-07
Update Justify Items example to remove duplicate lines

weaversam8:
This docs page has a few duplicate `<div>` tags that don't need to be there.


bradlc:
Thanks!

Issue #666, opened 2020-12-08
Missing JSX support in documentation


jonasdev:
![image](https://user-images.githubusercontent.com/26645731/101525053-1033ce00-398b-11eb-91a1-c766e287a093.png)

Spend hours, to figure out why styling is not working after build on Netlify. Then I added `'./src/**/*.jsx'` to purge array, which solved the problem. Should be mentioned in documentation, since .jsx is pretty common, and even a lint rule if using airbnb linter

PR #667, opened 2020-12-08
Add jsx to purge

jonasdev:
Will solve the issue I had https://github.com/tailwindlabs/tailwindcss.com/issues/666

### DIFF
--- a/src/pages/docs/guides/create-react-app.mdx
+++ b/src/pages/docs/guides/create-react-app.mdx
@@ -67,6 +67,8 @@ types:
 purge: 
   - ./src/**/*.js
   - ./src/**/*.jsx
+  - ./src/**/*.ts
+  - ./src/**/*.tsx
   - ./public/index.html
 ```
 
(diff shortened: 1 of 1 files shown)


bradlc:
Hey thanks @jonasdev! I think we should update this for sure. If you don't mind please could you update it to be a single glob, and add `ts` and `tsx` as well? If it can match the glob in the [Gatsby guide](https://github.com/tailwindlabs/tailwindcss.com/blob/master/src/pages/docs/guides/gatsby.mdx) that would be perfect 👍

jonasdev:
@bradlc Sure. Good idea. Didn't catch the glob in first commit, which caused another commit, my bad. It should be working now

bradlc:
Thanks! Sorry to be a pain, can we make this a single item? i.e. `./src/**/*.{js,jsx,ts,tsx}`

jonasdev:
Just the way you want 👌

bradlc:
Perfect, thank you! 🙌

PR #679, opened 2020-12-15
Update laravel.mdx

DavydeVries:
Changed title of the Laravel installation guide with the version number of laravel.


simonswiss:
Hey!

While it does add some context to specify the version of Laravel in the title - for consistency purposes with the other guides (which never mention the version of the framework in question), I think we should keep it to saying just `Laravel`.

These guides will be updated regularly in the future, so I think we can always assume that the guides are using the latest stable version of things.

Thanks for the suggestion, though! 👍

PR #681, opened 2020-12-15
fix typo: update `global.css` to `globals.css`

mvllow:
match next.js `globals.css` also mentioned in this guide


simonswiss:
Good pickup, thanks for this! 🙏

PR #685, opened 2020-12-16
Fix classes in inset-x example code

SEAPUNK:
In https://tailwindcss.com/docs/top-right-bottom-left, the classes in the top-right-bottom-left demonstration do not match the classes in the example code.

Setting the width for inset-x elements would not create the "span edge" effect.


bradlc:
Thanks!

PR #686, opened 2020-12-16
Update HTML comment in 'flex-none' example block

pboi20:
Hi! Just a quick fix that I noticed while browsing the documentation. I updated the example code block to indicate that flex-none items will not grow or shrink.


simonswiss:
Thank you for this 🙏

PR #687, opened 2020-12-17
Use default colors in gradient examples (teal -> green, orange -> yellow)

tchin25:
This is basically an extension of #625 where I spent an embarrassingly long amount of time trying to figure out why the code examples weren't working for me and it was because both teal and orange aren't default colors anymore.

### DIFF
--- a/src/pages/docs/background-image.mdx
+++ b/src/pages/docs/background-image.mdx
@@ -16,7 +16,7 @@ To give an element a linear gradient background, use one of the `bg-gradient-{di
 
 ```html
 <template preview>
-  <div class="h-24 bg-gradient-to-r from-yellow-400 via-red-500 to-pink-500"></div>
+  <div class="h-24 bg-gradient-to-r from-orange-400 via-red-500 to-pink-500"></div>
 </template>
 
 <div class="**bg-gradient-to-r from-yellow-400 via-red-500 to-pink-500** ..."></div>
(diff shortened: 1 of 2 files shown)


simonswiss:
Nice one!

In certain cases, we kinda "cheated" and used the _extended_ colors for the actual code example, but translated to the closest default color (example: emerald->green) for the HTML snippets, just so the examples looked a bit prettier.

It's not ideal to copy/paste something that will look different - but having really nice looking docs examples is also important.

For example, on this button, the contrast is a little too harsh with the "green to blue" instead of "teal to blue".

Would you mind reverting to the extended colors for the `preview` examples, but keep the changes you made for the code snippets?

src/pages/docs/background-image.mdx
```diff
@@ -16,10 +16,10 @@ To give an element a linear gradient background, use one of the `bg-gradient-{di
 
 ```html
 <template preview>
-  <div class="h-24 bg-gradient-to-r from-orange-400 via-red-500 to-pink-500"></div>
+  <div class="h-24 bg-gradient-to-r from-yellow-400 via-red-500 to-pink-500"></div>
```
simonswiss, line 19: Based on my comment in the PR - I would revert the colors inside the `<template preview>` tag, but not in the HTML right below (which is the code snippet).

```suggestion
  <div class="h-24 bg-gradient-to-r from-orange-400 via-red-500 to-pink-500"></div>
```

tchin25:
Sure no problem 👍

simonswiss:
Awesome - thanks for this 🎉

🚀

PR #690, opened 2020-12-17
Fix to refer dark mode docs

Mizumaki:
In Dark Mode section, fix to refer dark mode docs, instead of responsive design docs.


simonswiss:
Good catch! Thanks for this 👍

PR #707, opened 2021-01-06
fix(home/mobileFirst): remove extra top padding on sm screens

BenNG:
Hello ! 

When copy/pasting the code from the home/mobileFirst example, there is an extra top padding on sm screens

image

A way to fix the problem is adding `sm:pt-0` to the same element that has `pt-40`.

This PR contains this exact change

Cheers


adamwathan:
That padding is supposed to be there, with your change the demo looks like this:

image

BenNG:
This is weird... As I just copy/paste the code.

PR #711, opened 2021-01-11
Fix link in anchor

helmut:
Minor update because the link to adding tailwind as a PostCSS plugin within the anchor was broken.


helmut:
Apologies for not search for an existing pull request first. Closed.

PR #749, opened 2021-02-09
Fix outline focus color to ring-blue-500

notapatch:
- The website example didn't seem to do anything even when I
   removed focus:border-blue-400. The ring documentation suggested
   ring prefix instead of border. This seemed to work.
---

Note: It only looks the same as the default when I use blue-500 and focus:ring-opacity-50 in. But I did not want to make a change on something I didn't understand.


adamwathan:
Hey! This documentation section is just explaining the `outline-none` class, the `...` in the class list is implying that there are a bunch of other classes in place as well, so copying and pasting it is never going to actually give you the full result. If you view source on the real example you can see all of the other classes in place. The `focus:border-blue-300` class is correct:

image

PR #770, opened 2021-03-02
Update SVGs and object-fit examples to better show behaviour

27pchrisl:
The "placeholder" SVGs do not have a big enough initial viewport to show off the CSS rules being applied to them. This mainly affects object-fit examples and the "bg-auto" example.

Currently some of the object-fit examples look visually identical, which could confuse someone on which to use. For example compare "object-contain" and "object-fill", and also "object-none" and "object-scale-down".

The "bg-auto" example also does not really show what happens with a large image.

I have modified the placeholder SVGs to have a much larger initial size, so they are more likely to be resized where necessary. I have also added preserveAspectRatio="none" so that the browser is allowed to ignore the aspect ratio when rendering examples such as "object-fill".

I added a minor update to the object-fit explanations to indicate where aspect ratio is taken into account, and added a second "object-contain" example to show the content being either width or height contained.

The SVG change did have a visual effect change in the "background-repeat" examples, where I have explicitly set the background-size to 12rem to match the previous size.

I couldn't see any visual changes in the other places where these placeholders are used (clear, background-attachment, background-size (except bg-auto), float) as they all have their viewport width or height defined.

The "object-scale-down" example still looks like the second "object-contain" example, I'm not sure what the best way would be to represent visually what that rule does.


27pchrisl:
Vercel had not deployed the changes before, but it has now.

reinink:
Hey! Thanks so much for trying to help improve the Tailwind CSS docs 🙏 

However, we have since launched a new version of the website in December, and have come up with all new examples here that I think better explain `object-fit` and `background-repeat`:

https://tailwindcss.com/docs/object-fit
https://tailwindcss.com/docs/background-repeat

Let us know if you feel any further improvements could be made to these new examples 👍

PR #784, opened 2021-03-12
Fix line in Next.js guide for coding style consistency

feute:
This is a small change.

Basically, the coding style used in Next.js by default (and in the guide) is to use single quotes and without semicolons. The added import from the guide instructions is:

```js
import "tailwindcss/tailwind.css";
```

But, for consistency with the rest of the docs, it should be:

```js
import 'tailwindcss/tailwind.css'
```


simonswiss:
Nice one, thanks for this subtle but nice quality improvement PR 🙏

PR #797, opened 2021-04-05
Fixed typo in "Dark Mode" page

barnoma-soz:


simonswiss:
Thanks for that @barnoma-soz 🎉

PR #825, opened 2021-04-29
Add "env" keyword to docs for cross-platform .env

DoctorDerek:
Adding "env" before setting environmental variables makes it cross-platform (i.e. work on Windows in addition to Unix / Mac).

Tested Windows 10 x64

"@Infinity @jamie-penney env NODE_ENV=test mocha --reporter spec will use the declared environment variable in a natively cross platform fashion, but the key is it is used by npm in an ad hoc and one-time fashion, just for the npm script execution. (It's not set or exported for future reference.) As long as you're running your command from the npm script, there's no issue. Also, the "&&" must be removed when doing it this way. – estaples Apr 5 '18 at 16:10 " Source: https://stackoverflow.com/a/27090755

### DIFF
--- a/src/pages/docs/just-in-time-mode.mdx
+++ b/src/pages/docs/just-in-time-mode.mdx
@@ -86,9 +86,9 @@ By default, Tailwind will start a long-running watch process if `NODE_ENV=develo
   // ...
   scripts: {
     // Will start a long-running watch process
-    "dev": "NODE_ENV=development postcss tailwind.css -o ./dist/tailwind.css -w"
+    "dev": "env NODE_ENV=development postcss tailwind.css -o ./dist/tailwind.css -w"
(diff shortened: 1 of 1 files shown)


adamwathan:
Hey thanks for this! Unfortunately I think I'm going to keep it as-is, just because this is not a syntax I've ever seen before in the documentation for any other libraries and to me as a reader it feels surprising/confusing. The node docs themselves don't include anything like this in their examples either for example:

https://nodejs.org/api/cli.html#cli_node_options_options

If this keeps coming up as a common problem can definitely revisit but for now I'd like to stick the conventions used by other tools.

DoctorDerek:
No worries, I've never seen any Windows developer support from other tools
either 😅 And it is an odd syntax.

Thanks for taking a look, Adam!

PR #854, opened 2021-05-25
Add link to layers for further clarity

tim-andes:
New to Tailwind and was reviewing the docs for the first time. I spent a while trying to figure out what layers were after learning about extracting components. Couldn't find an explanation for a while, so a link would be helpful.


simonswiss:
Hey @tim-andes 👋

Thanks for this. It's always useful to get feedback from folks who are new to Tailwind and our documentation, always gives a good perspective on some of the pain points were are sometimes _too close to the metal_ to be able to perceive.

I think this change makes sense - it adds some useful context without really changing anything else, so I'm going to merge it!

Thanks for your contribution 🎉

PR #870, opened 2021-06-16
Update docs for Preprocessors

depsimon:
Add `-D` flag for dev dependencies.
Prevent error with `require()` in `postcss.config.js` cfr [this issue](https://github.com/vercel/next.js/issues/10117)


src/pages/docs/using-with-preprocessors.mdx
```diff
@@ -130,10 +130,10 @@ It's included directly in the `tailwindcss` package itself, so to use it all you
 // postcss.config.js
 module.exports = {
   plugins: [
```
vltansky, line 132: I think it should be object `plugins: {}`

simonswiss:
Better late than never - thanks for that 👍

I've updated the arrays to be objects instead to conform to the new format.

PR #878, opened 2021-06-21
Add missing documentation for :empty pseudo-class

e-krebs:
Which was added with [this PR](https://github.com/tailwindlabs/tailwindcss/pull/3298)


reinink:
Hey! Thanks so much for this contribution, but we've since launched a whole new version of the website, which includes a new section about this modifier: https://tailwindcss.com/docs/hover-focus-and-other-states#empty

Thanks either way! 🙏

PR #889, opened 2021-07-04
Info about negative background position values

FrankyFrankFrank:
While negative values are possible in the background position configuration, the documentation does not yet mention this possibility.
This PR includes some helpful information to anybody who is interested.

### DIFF
--- a/src/pages/docs/background-position.mdx
+++ b/src/pages/docs/background-position.mdx
@@ -113,6 +113,25 @@ By default, Tailwind provides nine `background-position` utilities. You change, 
   }
 ```
 
+### Negative values
+
+If you'd like to add negative background position classes (taking the form `-background-{direction}-{size}`), prefix the keys in your config file with a dash:
+
+```diff-js
+  // tailwind.config.js
+  module.exports = {
+    theme: {
+      extend: {
+        backgroundPosition: {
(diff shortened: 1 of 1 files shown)


reinink:
Hey @FrankyFrankFrank — thanks so much for this contribution. Since you submitted this we've released Tailwind CSS v3.0, which has a few changes that are relevant here.

First, now when you append a dash (`-`) to any class, Tailwind automatically attempts to make the value negative — even if you don't include a negative value in your config. This works great for things like margin and the top/left/bottom/right utilities.

This, however, doesn't work well with the background position utilities as they can contain words and Tailwind can't auto convert them to negatives. You can, of course still prefix a utility name with a dash in your config, and Tailwind will automatically move the dash to the very beginning of the class name. However it doesn't do anything fancy beyond that.

Now, while this is maybe helpful information to know, practically speaking we haven't really been having people reporting issues with this, so I'm inclined to just leave this out of the docs for now. If it becomes an issue, we can always revisit.

It's also worth noting that v3.0 comes with arbitrary value support, which is a really nice way to handle custom background position values without having to edit your config at all:

```html
<div class="bg-[left_-18rem_center]" style="background-image:url(https://...)">
  <!--  -->
</div>
```

Hope that makes sense — and thanks either way for the contribution 👍

PR #891, opened 2021-07-08
Extend limitations for space utils about divide

tordans:
It is not possible to use `space-*` and `divide-*` together in a sensible ways which is a limitation that is well fitted to be part of the docs.

Ping https://github.com/tailwindlabs/tailwindcss/discussions/2466

### DIFF
--- a/src/pages/docs/space.mdx
+++ b/src/pages/docs/space.mdx
@@ -102,7 +102,7 @@ For those situations, it's better to use the [gap utilities](https://tailwindcss
 
 ### Cannot be paired with divide utilities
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/space.mdx
```diff
@@ -100,6 +102,10 @@ For those situations, it's better to use the [gap utilities](https://tailwindcss
 </div>
 ```
 
+### Divide utilities
+
+The `space-` utilities are not designed to work together with the [`divide-` utilities](https://tailwindcss.com/docs/divide-width). For those situations, consider adding maring/padding utilities to the children.
```
DoctorDerek, line 107: margin

simonswiss:
Thanks for this - it's indeed a common "pitfall" and I think that has its place in the documentation 👍

PR #901, opened 2021-07-23
Rename animation timing function

jonassiewertsen:
I think the CSS Property should be renamed to `animation-timing-function` in the docs regarding the MDN Web documentation:

https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timing-function


simonswiss:
Hey! Good catch, that's correct - this should not show the CSS-in-JS camel casing, since we're showing the CSS output in that table 👍

Thanks for this!

jonassiewertsen:
Thanks for the fast merge and nice that could participate with my first little PR to the TailwindCSS docs. 

Amazing work!

PR #909, opened 2021-08-07
Update installation.mdx

aesfer:
There was a missing step I had to do to get the compatibility build working so it may help others if you add it to the docs.

Cheers,
Andres

### DIFF
--- a/src/pages/docs/installation.mdx
+++ b/src/pages/docs/installation.mdx
@@ -387,6 +387,18 @@ npm uninstall tailwindcss postcss autoprefixer
 npm install -D tailwindcss@npm:@tailwindcss/postcss7-compat postcss@^7 autoprefixer@^9
 ```
 
+And be sure to update the plugin name inside `postcss.config.js` like this:
+
+```js
+// postcss.config.js
+module.exports = {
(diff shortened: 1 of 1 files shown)


huibizhang:
Is it necessary ?

I'm using compatibility build for vue-cli and creat-react-app, but I don't need to change plugin name to `@tailwindcss/postcss7-compat` in `postcss.config.js`

Just only use `tailwindcss` is okay for me.

Because in package.json , it looks like:
```diff
"devDependencies": { 
    "autoprefixer": "^9.8.6",
    "postcss": "^7.0.36", 
+   "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.2.7" }
```

aesfer:
I installed it in my project by manually adding it inside devDependencies like this:

```
  "devDependencies": {
    "@tailwindcss/postcss7-compat": "^2.2.7",
```

But since the instructions in the docs say to install it via `npm install -D tailwindcss@npm:@tailwindcss/postcss7-compat`, the result should be like yours. Therefore this is not only unnecessary but it would actually break the build.

I'll close this PR, thanks for noticing :)

PR #916, opened 2021-08-16
Take out line where global.css is removed

dgrcode:
In the section "Import Tailwind directly in your JS", the first code block indicates to remove `import '../styles/globals.css'`. Later on the same section, there's a note to "ensure your CSS file is being imported" in the same place we removed it earlier.

### DIFF
--- a/src/pages/docs/guides/nextjs.mdx
+++ b/src/pages/docs/guides/nextjs.mdx
@@ -37,7 +37,6 @@ If you aren't planning to write any custom CSS in your project, the fastest way 
 
 ```diff-js
   // pages/_app.js
-- import '../styles/globals.css'
 + import 'tailwindcss/tailwind.css'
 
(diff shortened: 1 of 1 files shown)


adamwathan:
Hey! This is deliberate, you remove `globals.css` and import `tailwindcss/tailwind.css` instead 👍🏻 

The second section that references `globals.css` is showing an alternative method:

image

So it's saying there are two options, option 1 is "Import Tailwind directly in your JS" and option 2 is "Include Tailwind in your CSS". You aren't meant to do both steps, just one 👍🏻 

Hope that helps.

dgrcode:
Woops, I definitely didn't read it thoroughly enough. I understood (wrongly from my skim-reading) that each code block was describing a step, so I actually did both. Since it worked I figured removing the `global.css` line was extra work that was being undone later.

Completely understood now. Sorry for wasting your valuable time 🙏

PR #950, opened 2021-09-17
Fix multiple escapes in plugin code example

marcushjelm:
Minor fix to "supports-grid" plugin example to avoid multiple escapes of content in `${rule.selector.slice(1)}`.

**Example**:

`class="supports-grid:hover:underline"`

Before:
```
@supports (display: grid) {
  .supports-grid\:hover\\\:underline\:hover {
    text-decoration: underline;
  }
}
```

After:
```
@supports (display: grid) {
  .supports-grid\:hover\:underline:hover {
    text-decoration: underline;
  }
}
```


reinink:
@marcushjelm Oh nice catch! Thanks for submitting this fix 💪

PR #965, opened 2021-10-13
feat: show px unit preview for rem units

### LINKED ISSUE #946: `px` equivalent besides `rem` in docs
I like the Tailwind Intellisense because it gives me `px` styles now besides `rem` so I don't have to do the math.

But when I open `tailwind.config.js` to use `theme('borderRadius.md')` then I don't understand its value in `px` as Intellisense doesn't work there yet.

So I open docs & there it is in `rem` again. Would love to have `px` values as I try to copy CSS from online & then convert it to `px` but I don't know if I need to use JIT mode or the equivalent `px` value exists in `rem`.

A comment would be helpful like `rounded-md => border-radius: 0.375rem; /* 6px */`

iamandrewluca:
- [x] Pixel preview added to all values that end with `rem`
- [x] Pixel preview not added to values that end with `em`
- [x] Pixel preview not added to complex values e.g. `calc(1.25rem * var(--tw-space-y-reverse));`
- [ ] ~Find a way to allow user to chose own html font size~

<details>

<summary>Images</summary>

image

image

image

</details>

Closes #946

<details>
<summary>Meme</summary>

image

</details>


iamandrewluca:
@adamwathan @bradlc what are your thoughts on this?

adamwathan:
Hey! Appreciate the contribution and definitely a cool idea — right now I'm afraid I just don't have the available attention to think through exactly how I'd want this to look or work, so it could be many weeks before I have a chance to make this a priority, especially since I would need to pull in someone from the design team to look as well.

Thanks again though — hope it's understandable that this is probably going to sit for a while as I've just got lots on my plate already and can't realistically cancel those plans to work on ideas that come in as PRs of course, especially when they take a lot of thinking ❤️

iamandrewluca:
Sure 🙂 no rush. Thanks for response!

reinink:
@iamandrewluca Hey thanks for all your work on this, and for keeping this PR up-to-date, even over the migration to the new v3 docs 👏

So I was just chatting with the team about this today, and I think we want to pull the trigger on this. I'm going to review your PR in more detail tomorrow.

Also, I don't think we need to add a way for users to choose their own font size. I think assuming a root font size of `16px` is totally fine for the purposes of the documentation.

iamandrewluca:
@reinink cool, then I'm updating this PR one last time before your review tomorrow!

reinink:
@iamandrewluca Hey! I hope you're okay with this, but we ended up making a bunch of changes to your implementation. We basically updated our existing `ClassTable` to automatically look for any `rem` values, and when it finds them automatically spit out the corresponding `px` value. This is nice as we don't have to many any changes to the individual pages — it just all happens automatically.

In addition to that, we also tweaked the design a little. Instead of adding these as a third column, we instead put them as comments right beside the values in the "properties" column. Same basic idea though!

image

Thanks so much for all the leg work you've done on this — it's great to finally have this merged in! 🥳

iamandrewluca:
@reinink Thanks! That's event better, it looks more similar with vscode extension 🚀 🎉

PR #979, opened 2021-11-02
Clarify when this class can be used

heymartinadams:
Only figured this one out once I came across [Adam’s comment](https://github.com/tailwindlabs/tailwindcss/discussions/2533#discussioncomment-94170). Would be helpful to add it to the docs to save time.

### DIFF
--- a/src/pages/docs/fill.mdx
+++ b/src/pages/docs/fill.mdx
@@ -4,6 +4,7 @@ description: "Utilities for styling the fill of SVG elements."
 ---
 
 import plugin from 'tailwindcss/lib/plugins/fill'
+import { TipGood, TipBad } from '@/components/Tip'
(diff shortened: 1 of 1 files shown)


reinink:
Hey, thanks for this contribution!

While I agree that this is a bit of a confusing thing with SVGs that have been implemented as images, I'm not sure that this makes sense for the Tailwind CSS documentation as it's not technically a Tailwind issue per se, but more just a known limitation of working with SVGs this way.

The more annoying problem is that if we did add it to the fill page, we technically need to also add it to the stroke and stroke width pages as well.

That said, we have talked about eventually adding a dedicated SVG guide to the docs, and I think that would be a great place for this tip.

I am going to close this PR, but we will keep this in mind should we ever add that guide.

Thanks again either way! 🙏

PR #981, opened 2021-11-09
postcss.config.js syntax has changed

TheHippo:
The `postcss.config.js` syntax has changed and the documentation does not match the syntax that is generated by `tailwindcss init --postcss`.


TheHippo:
Either I'm bad at searching or the Github search has failed me, but #870 should be merged instead.

PR #994, opened 2021-12-04
Update dark-mode.mdx

Lzzzs:
I think adding this line can explain why it can be more detailed, because I have been wondering why my dark mode does not work these days. It is because I configured the prefix, but I see that the official website does not say that if the prefix is configured,dark mode must be prefixed to work, I think this can make people less to step on the hole


reinink:
Hey thanks for this suggestion! 🙏 

Since you submitted this we've actually launched a whole new version of the website, and since I wanted to make some tweaks to what you started, I just did this as a separate commit and marked you as a contributor: 20d2648e5c314d0ea777c2781b555294e8a55e10

PR #998, opened 2021-12-09
Fix reference to safelist option as top-level configuration

metaturso:
This PR fixes a tiny inconsistency between the documentation text and its corresponding snippet of code. 

The `safelist` option should live under the `purge` option, not in the top-level configuration.


bradlc:
Thank you!

PR #1003, opened 2021-12-09
Add emacs to the editor support page

spcbfr:


reinink:
Hey thanks for this contribution! While we totally understand that there is ways to make Tailwind CSS work with other editors like Emacs and Neovim, but for now we want to keep this page limited to just the official plugins available. That includes our Intellisense plugin for VS Code, as well as the first-class support provided by JetBrains.

The primary reason for this is because we know that these two plugins are actively maintained and supported. However we can't verify that for other editors.

Hope that makes sense, and thanks either way! 🙏

PR #1009, opened 2021-12-09
Fixing valid/invalid variant docs

jaydiablo:
Noticed that the valid/invalid variants were copied from required and weren't changed.


simonswiss:
Great pickup, thank you!

PR #1017, opened 2021-12-10
Update customizing-colors.mdx

willemvb:
Update 2 links in introduction with correct anchor names.


reinink:
Good catch, thanks @willemvb! 🙏

PR #1021, opened 2021-12-10
Fix underline dotted, wavy example in v3 blog

bhatvikrant:


reinink:
Good catch, thanks! 🙏

PR #1022, opened 2021-12-10
Use plural in upgrade guide

NoelDeMartin:
Indeed I wasn't affected by the change, and no sharks in sight so far :crossed_fingers:.

But I think it's better to use the plural here to be congruent with the rest of the text.

Great job with the guide by the way, it was a really easy upgrade :D.


reinink:
Heh, thanks for the suggestion, and it's technically more correct, but I think we're going to leave this as is, since it suggests that this is a fun comment from Adam 😁🦈

PR #1050, opened 2021-12-14
Add ts,tsx to guides/create-react-app.js

JakobLyon:
I was stuck on an issue for about an hour where my typescript files weren't getting styles and this was the culprit. It's a little obvious and I should have seen it coming, but mentioning it in the docs might help the next person using typescript and trying to add tailwind.


reinink:
Thanks, good suggestion 👍

Note, I simplified it a little — I didn't think the extra explanation was necessary, and we've tried really hard to keep the descriptions on the left as terse as possible.

PR #1061, opened 2021-12-17
Add minor change in upgrade guide for `current` key when overriding theme `colors`.

wking-io:
This issue: https://github.com/tailwindlabs/tailwindcss/issues/6371

Helped me solve an "bug" I had with fill-current not working. Wanted to add to upgrade guide so that others can find this and not need to look into the issues to solve.


wking-io:
No worries! 👌

PR #1066, opened 2021-12-19
Prevent page reload when hitting Enter in input

laundy:
When hitting enter while being in the `Filter Projects...` input, the whole page reloads and does not scroll back to the position the user was before. IMHO this is bad for UX.


reinink:
Hey! Thanks for this idea, but I think the better way to solve this is to just replace the `<form>` with a `<div>` to avoid this behavior to begin with. I've made this change (61a756787875efebc283be001c7ea7c38197b5e9), and marked you as a coauthor of that commit 👍

PR #1067, opened 2021-12-19
add the -p option for installation instructions for postcss

crswll:
Someone pointed this out in Discord and it makes sense to me. It won't touch an existing `postcss.config.js` file and makes the next step *way* easier.

### DIFF
--- a/src/pages/docs/installation/using-postcss.js
+++ b/src/pages/docs/installation/using-postcss.js
@@ -15,7 +15,7 @@ let steps = [
     code: {
       name: 'Terminal',
       lang: 'terminal',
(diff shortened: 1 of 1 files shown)


reinink:
Hey! Thanks for this, but we actually intentionally left this flag out of this guide since some tools recommend configuring PostCSS in different configuration files.

For example, Nuxt.js recommends adding this to your `nuxt.config.js` file, and Laravel recommends adding it to your `webpack.mix.js` file.

Our hope with the current guide is that folks will simply update their PostCSS configuration wherever that is in their framework. If anything, I might actually update the second step to make that more clear.

Thanks anyway! 👍

crswll:
No problem! Also, good to know those other details, thanks!

PR #1072, opened 2021-12-20
Add missing step

arange:
I followed the exact same steps on this page but failed to enable tailwind with nextjs until I find this missing file in the example, also global style is mentioned in nextjs's tutorial: https://nextjs.org/learn/basics/assets-metadata-css/global-styles

### DIFF
--- a/src/pages/docs/guides/nextjs.js
+++ b/src/pages/docs/guides/nextjs.js
@@ -69,6 +69,26 @@ let steps = [
     },
   },
   {
+    title: 'Enable global styles',
+    body: () => (
+      <p>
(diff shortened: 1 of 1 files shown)


reinink:
Hey, thanks for this but the default Next.js app created with `create-next-app` comes with a `pages/_app.js` file which imports the `globals.css` file. You can see that [here](https://github.com/vercel/next.js/blob/0af3b526408bae26d6b3f8cab75c4229998bf7cb/packages/create-next-app/templates/default/pages/_app.js#L1).

Because of this, and in the interest of keeping the guides as short as possible, I don't think adding this step makes sense.

Thanks either way! 🙏

PR #1073, opened 2021-12-20
Fixed "Open/closed state example" with Dark Mode

NovusOpus:
In the [Open/closed state](https://tailwindcss.com/docs/hover-focus-and-other-states#open-closed-state) The `summary` text is almost invisible while in closed state with Dark mode. 

Current view:
Screenshot (683)

Added some `dark:` utilities so it can work with Dark mode.


reinink:
Nice catch, this looks great 👍

PR #1089, opened 2021-12-24
Added aria-label for accessibility

DavidIsrawi:
**Issue**
- Progress Bar example in the music player component was missing an aria-label for screen readers to leverage (this was flagged by a Lighthouse accessibility run)

**Fix**
Added aria-label to DIV


DavidIsrawi:
CC: @adamwathan

bradlc:
Thank you!

PR #1090, opened 2021-12-25
Fix Doc Search icon repeats and look odd

Nivethan-Ar:
Hi 👋

when searching for something on the documentation page  that search icon is the only thing that looks something odd. 
image

so i checked in the developer tools and it shows the normal search icon as in the following image.
image

so i guessed the second image get repeated that creates a weird look. so i removed the background repeat and it looks cool now!

Thanks


bradlc:
Hey @Nivethan-Ar. Which browser and operating system are you seeing this in? I can't reproduce the issue but I think a better solution would be to set the `background-size` instead, otherwise the icon is not the correct size:

```diff
- background-repeat: no-repeat round;
+ background-size: 100% 100%;
```

Nivethan-Ar:
> Hey @Nivethan-Ar. Which browser and operating system are you seeing this in? I can't reproduce the issue but I think a better solution would be to set the `background-size` instead, otherwise the icon is not the correct size:

i use brave browser and firefox, but on firefox it looks normal, i get this issue only on brave browser.  can i change the background size?

bradlc:
Interesting, I can't reproduce in Brave on macOS or Windows, but if you change this PR to use `background-size: 100% 100%` instead of `background-repeat` and verify that it resolves the issue for you I'll happily merge it 👍

reinink:
Hey, closing this one for now since we are unable to reproduce this ourselves, and this PR has been sitting here stale. Feel free to resubmit a PR with the corrections that @bradlc suggested 👍

PR #1094, opened 2021-12-28
Update package.json

foucist:
Increase browser support from 85% to 90% coverage

> npx browserslist --coverage "> 1%"
> These browsers account for 85.27% of all users globally
> npx browserslist --coverage "> 0.5%"
> These browsers account for 90.21% of all users globally

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "format": "prettier '{src,remark,scripts}/**/*.{css,js,mdx}' --write"
   },
   "browserslist": [
-    "> 1%"
+    "> 0.5%"
   ],
   "dependencies": {
     "@badrap/bar-of-progress": "^0.1.1",


foucist:
Based on the preview app produced by the CI, the change doesn't fix the bug I'm encountering in Safari 13.1.

v3 site refuses to display properly.  It throws a next js client side error message and a bunch of these types of messages:

`[Error] TypeError: b.addEventListener is not a function. (In 'b.addEventListener("change",S)', 'b.addEventListener' is undefined)`

Any ideas on source of error?

reinink:
Hey thanks for reporting this bug! I was able to reproduce to this and fix it: 6024007d1578e14109509a154b89661762064914 👍

foucist:
interesting, so it was due to using addEventListener instead of addListener which is deprecated

reinink:
Yeah, Safari 13 doesn't have support for the new `addEventListener` method, so we had to detect that and use the older `addListener` method instead.

Issue #1099, opened 2021-12-30
Docs: Tailwind CSS configuration with Gatsby is missing some information


ashiishme:
The installation guide of **Tailwind CSS** with **Gatsby** framework is missing some information that is required when someone creates a `Gatsby theme` with Tailwind CSS.

After following the installation guide, when running the project the build fails because Gatsby is not able to find the `tailwind.config.js` file in the project that depends on the Gatsby theme build with Tailwind CSS.

**To reproduce:**
- Create a test Gatsby theme and configure Tailwind CSS as per [the documentation](https://tailwindcss.com/docs/guides/gatsby).
- Install the theme in a new Gatsby site
- Start the server

**Error:**

```ts
ERROR in ../<theme-directory>/src/styles/global.css
Module build failed (from ../node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from 
../node_modules/gatsby-plugin-postcss/node_modules/postcss-loader/dist/cjs.js):
TypeError: Cannot read properties of undefined (reading 'config')
    at getTailwindConfig
    ....
```

To fix this, we need to include the `tailwind.config.js` file in `postcss.config.js`.

I am interested to update the documentation with the relevant information.

PR #1100, opened 2021-12-30
docs: update installation guide with Gatsby

ashiishme:
#1099: Updates installation guide with Gatsby.

### DIFF
--- a/src/pages/docs/guides/gatsby.js
+++ b/src/pages/docs/guides/gatsby.js
@@ -61,6 +61,26 @@ let steps = [
     },
   },
   {
+    title: 'Configure PostCSS for Gatsby theme ( optional ) ',
+    body: () => (
+      <p>
+        In your <code>postcss.config.js</code> file, include Tailwind CSS configuration file path if
+        you are building a Gatsby Theme with Tailwind CSS.
+      </p>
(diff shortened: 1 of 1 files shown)


reinink:
Hey there! Thanks so much for taking time out of your day to try and improve the Tailwind CSS docs 👍

Unfortunately, while I totally get that there is an extra installation step required to make Tailwind CSS work with Gatsby themes, our goal with the framework guides is to keep them *as simple as possible*. Meaning, we don't include any optional steps, even though for each framework there are lots of them available.

I am going to go ahead and close this PR and the related issue. Hope that makes sense, and thanks again either way! 🙏

ashiishme:
> Hey there! Thanks so much for taking time out of your day to try and improve the Tailwind CSS docs 👍
> 
> Unfortunately, while I totally get that there is an extra installation step required to make Tailwind CSS work with Gatsby themes, our goal with the framework guides is to keep them _as simple as possible_. Meaning, we don't include any optional steps, even though for each framework there are lots of them available.
> 
> I am going to go ahead and close this PR and the related issue. Hope that makes sense, and thanks again either way! 🙏

Agreed & totally make sense. Thank you for your time.

PR #1106, opened 2022-01-05
Docs: Fixes title used with map example

brwolfgang:
Just fixes the duplicate Team title


reinink:
Nice catch, thanks! 🙏

PR #1115, opened 2022-01-06
Fixes typo by adding the missing 'process.env'

ghoshnirmalya:
## Details
This PR adds the missing "process.env" for the optimizing section.


reinink:
Good catch, thanks! 💪

PR #1119, opened 2022-01-09
Add Ruby on Rails Framework Guide

99darshan:


thecrypticace:
Thanks for the PR! 🎉  It served as an excellent starting point. I made some tweaks to simplify things and more closely match our existing guides.

Much appreciated!

PR #1123, opened 2022-01-10
Fix: Docs `cursor-none` image

marcbouchenoire:
This tiny PR is a one-file change which fixes the `cursor-none` image example—it’s currently displaying as `not-allowed` instead of nothing.


reinink:
Hey good catch! I've tweaked this implementation slightly to just remove the image entirely for `cursor-none`. Thanks for reporting this 🙏

PR #1125, opened 2022-01-11
Fix typo in doc (remove extra word)

julien-me:


reinink:
Good catch, thanks! 🙏

PR #1129, opened 2022-01-11
Update docs for Flex Basis chapter

yuxiaoy1:
The `flex-basis-[14.2857143%]` won't work, it should be `basis-[14.2857143%]`.


reinink:
Good catch, thanks! 🙏

PR #1132, opened 2022-01-13
Update link to raw Preflight CSS

MichaelAllenWarner:
In `preflight.mdx`, update the link to the raw Preflight CSS file (was for v2 of Tailwind, now is for v3).


simonswiss:
Thanks for that, @MichaelAllenWarner 👍

PR #1136, opened 2022-01-14
Update font-family.mdx

nyedidikeke:
Highlight, for consistency in the documentation, the fact that the three font family utilities that Tailwind provides by default comprise of
- a cross-browser sans-serif stack,
- a cross-browser serif stack, and
- a cross-browser monospaced stack.

Reference: https://tailwindcss.com/docs/font-family#customizing-your-theme

### DIFF
--- a/src/pages/docs/font-family.mdx
+++ b/src/pages/docs/font-family.mdx
@@ -14,7 +14,7 @@ export const classes = { utilities }
 
 ### Setting the font family
 
-You can control the typeface of text using the font family utilities.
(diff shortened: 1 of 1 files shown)


jacobthesheep:
Honestly don't know what the point of this is for. It's pretty easy to figure that out by looking at the fonts themselves

reinink:
Hey, thanks for this contribution and trying to make the Tailwind CSS docs better, but I don't really think this change is necessary, as it's clear to see in the class reference what fonts are included in each utility. Thanks anyway! 🙏

PR #1140, opened 2022-01-15
fix: renaming purge to content on presets doc

marceloadsj:
Small fix to update **purge** to **content** at https://tailwindcss.com/docs/presets#merging-logic-in-depth.


reinink:
Thanks again! 👍

PR #1141, opened 2022-01-15
fix: anchor for merging logic at presets doc

marceloadsj:
Small fix for anchor at end of https://tailwindcss.com/docs/presets#creating-a-preset section.


reinink:
Nice, thanks! 🙏

PR #1147, opened 2022-01-18
Add `npm install` step after init on Vite guide

simonswiss:
When creating a new vite app with `npm init vite xxx`, the dependencies are not installed and this is a required step.

### DIFF
--- a/src/pages/docs/guides/vite.js
+++ b/src/pages/docs/guides/vite.js
@@ -18,7 +18,7 @@ let steps = [
     code: {
       name: 'Terminal',
       lang: 'terminal',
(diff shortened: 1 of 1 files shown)


thecrypticace:
npm install is present in the next step and will install them:

Screen Shot 2022-01-17 at 19 43 09

Are you seeing different behavior?

simonswiss:
Hmmm, you can safely ignore me – I assumed that you needed to explicitly run `npm install` before installing specific packages, but looks like the `npm i -D tailwindcss postcss autoprefixer` also installs the other dependencies from `package.json` 🙊

Closing this!

PR #1150, opened 2022-01-18
Update angular.js

M0hammedImran:
Adding `autoprefixer` and `postcss` as peer dependencies.
`tailwindcss` does not work without them.

### DIFF
--- a/src/pages/docs/guides/angular.js
+++ b/src/pages/docs/guides/angular.js
@@ -28,7 +28,7 @@ let steps = [
     code: {
       name: 'Terminal',
       lang: 'terminal',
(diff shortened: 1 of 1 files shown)


reinink:
Thanks! Technically the latest version of Angular ships with both autoprefixer and PostCSS, so this isn't necessary for new projects. However, I'm not sure about older Angular projects, so this change is probably a good idea — especially since it doesn't hurt to add it as a dependency even if it's already one of Angular's dependencies 👍

M0hammedImran:
I didn't know about angular shipping with those dependencies. But few days back I bootstrapped an angular app with latest version of ng, tailwind wasn't working until I add PostCSS and autoprefixer.

PR #1164, opened 2022-01-25
Code snippet should be a self-closing tag

jakeb1090:
Img tag on line 3 will throw  error: Expected closing tag "div" to match opening tag "img"

### DIFF
--- a/src/pages/docs/utility-first.mdx
+++ b/src/pages/docs/utility-first.mdx
@@ -90,7 +90,7 @@ With Tailwind, you style elements by applying pre-existing classes directly in y
 ```html
 <div class="p-6 max-w-sm mx-auto bg-white rounded-xl shadow-lg flex items-center space-x-4">
   <div class="shrink-0">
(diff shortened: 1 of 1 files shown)


reinink:
Hey! As per the HTML specs this is totally valid syntax: https://html.spec.whatwg.org/multipage/syntax.html#start-tags

> 6. Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single U+002F SOLIDUS character (/). This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

In this case, `img` is a [void element](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), which means the `/` has no effect on it.

That said, if you're copying and pasting this into React (JSX) you'll run into issues, as all tags must be closed: https://react-cn.github.io/react/tips/self-closing-tag.html

Since this isn't intended to be a React example, but rather just an HTML example, I think this is fine as it is.

Thanks anyway! 👍

PR #1165, opened 2022-01-26
Mention how custom classes should be made.

iHildy:
You can't use Tailwind style class names for states like `hover:` instead, use something like `hover-` to avoid issues. The docs should mention this, <TipBad> might be a good idea.

An example of what I mean: https://play.tailwindcss.com/y6q8geyDzX

### DIFF
--- a/src/pages/docs/adding-custom-styles.mdx
+++ b/src/pages/docs/adding-custom-styles.mdx
@@ -265,7 +265,7 @@ Use the [`theme`](/docs/functions-and-directives#theme) function or [`@apply`](/
 
 ### Adding component classes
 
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

We already talk about this in this section: https://tailwindcss.com/docs/adding-custom-styles#using-modifiers-with-custom-css

> Any custom styles you add to Tailwind with `@layer` will automatically support Tailwind’s modifier syntax for handling things like hover states, responsive breakpoints, dark mode, and more.


I'll check with the rest of the team if we want to further improve this or not.

PR #1169, opened 2022-01-28
chore(postcss): fix code example for using postcss

Maybach91:
As mentioned in the text below, the package is called `postcss-nested` and *not* `postcss-nesting`.
 
`postcss-nesting` is not the official/standard package to include, so it would be maybe less confusing when copy the example code, which mentions a not so common package (postcss-nesting).


thecrypticace:
Hey,  thanks for the PR!

Understood that the `postcss-nested` plugin is the more common of the two — hence why it's the default when not passing in any plugin. However, some users may prefer to use something closer the work-in-progress CSS nesting specification. Because of this we also support using `postcss-nesting` which is what that section is intended to explain. Changing this would invalidate the instructions for those users since it is specifically about how to use the `postcss-nesting` plugin.

PR #1172, opened 2022-01-29
📝defaultClass to grid-template-columns arbitrary values

fauzulkc:
the arbitrary values should not be working if the class list doesn't contain a grid class. To avoid confusion a defaultClass="grid" should help mitigate the problem.


reinink:
Thanks, good catch! 👍

PR #1174, opened 2022-01-29
9 default font weight instead of 10

denniskloyn:


reinink:
Thanks, good catch! 👍

PR #1178, opened 2022-02-01
Replace text-normal with text-base

andreaselia:
As far as I'm aware, `text-normal` is not a thing, so I replaced it with `text-base`.


reinink:
Hey thanks for this! I think we can just remove `text-normal` entire — we don't need `text-base`. I've already updated it 👍

PR #1214, opened 2022-02-25
Document new color functions

thecrypticace:
Updated docs to go along with https://github.com/tailwindlabs/tailwindcss/pull/7665


thecrypticace:
We went with a different approach in the end — closing

PR #1220, opened 2022-03-12
Grid customizing should include how to use it

notapatch:
Currently customizing your theme tells you how to create a theme it does not instruct you how to use it. This commit would change that.

### DIFF
--- a/src/pages/docs/grid-template-columns.mdx
+++ b/src/pages/docs/grid-template-columns.mdx
@@ -75,6 +75,7 @@ You have direct access to the `grid-template-columns` CSS property here so you c
     }
   }
 ```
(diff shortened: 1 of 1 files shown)


reinink:
@notapatch Hey thanks for submitting this contribution and for your interest in Tailwind CSS!

So while I see how this might be helpful, if we made this change on this page, we should probably make this exact same change on every page in the docs that has a "Customizing your theme" section, such as the [margins](https://tailwindcss.com/docs/margin#customizing-your-theme) and [columns](https://tailwindcss.com/docs/columns#customizing-your-theme) pages.

Generally folks seem to be able to figure this out, so I think we'll just leave things as they are here.

I do appreciate your attempt to make our docs better though! 👍

PR #1268, opened 2022-05-23
Fixed typo: example code using wrong color class

timschonborn:
In the example that is being displayed the `bg-slate-800` class is used, see line 34.
However, the provided code snippet used `bg-slate-900`, see line 52.


thecrypticace:
Ooh good catch, thanks! 🎉

PR #1283, opened 2022-06-16
Add types to tailwind config in installation guide

sitek94:
Hey guys, amazing job with `3.1`! 🎉 

What do you think about adding types to `tailwind.config.js` in installation guide, since they're now included by default? 

Screenshot 2022-06-16 at 06 20 43


sitek94:
Happy to help, cheers! 🥳

PR #1285, opened 2022-06-17
feat: add type config on examples for vite, remix and nextjs

veritem:


thecrypticace:
Hey, thank you for the PR. We decided that we wanted these for the entire installation guide and there was already a PR open for that so I made some tweaks and merged in that PR: #1283

In any case, thank you for the contribution!

PR #1131, opened 2022-01-13
Add Symfony framework guide

AmProsius:
I added a Symfony framework guide similar to the Laravel one. You can find a working repository of this guide [here](https://github.com/AmProsius/symfony-tailwind). I created that repository as a temporary solution and will probably remove it if this guide will be merged.

PR #1286, opened 2022-06-21
Symfony added to the installation documentation

julienallexandre:


thecrypticace:
Hey, really appreciate the PR! I opted to use the original PR we had for Symfony (#1131) as a starting point however I took some details from this one as well as they both had some rather good things to pick from. I marked you as a co-author on the PR. Really appreciate it. Goal is to get three guides out today — as such they're going into a release branch / PR so we can release them all at once: #1326

PR #1289, opened 2022-06-29
feat: updated laravel docs to vite

dacoto97:


thecrypticace:
We added the Laravel Mix guide back for older, existing projects but put it under a separate tab. Thanks again @dacoto97!

PR #1291, opened 2022-07-01
Typo Fix

niclm:
Step 5 in the Vite installation guide suggests importing the newly-created `./src/index.css` file but the code snippet is missing the `src` directory.


thecrypticace:
Hey! So, in this case the relative import is present in the `./src/main.js` file. This means that the path `./index.css` is correct as it's in the same directory as the `main.js` file.

PR #1297, opened 2022-07-07
Guide people toward using path resolution.

nathanhammond:
In a workspace setup encouraging people to directly path into `node_modules` will result in inconsistent outcomes. Instead rely on the system to properly identify the path of the node module, and then append the glob pattern manually.

This will result in a system-path-separator normalized absolute path which is distinct from the current `contents` items. I did not investigate if that will work as expected where these values are consumed so I wanted to draw attention to the distinction.


thecrypticace:
Thanks for the PR. I made some adjustments to work better with the flow of the documentation. Really appreciate the starting point! ✨

PR #1311, opened 2022-07-27
docs: replace unescaped font names used in sample

xuatz:
I was looking at the docs regarding escaping font names.

The docs said that:
```
// Won't work:
'sans': ['Exo 2', ...],
```

But later on has this in the sample code snippet:
```
module.exports = {
  theme: {
    extend: {
      fontFamily: {
+        'sans': ['Proxima Nova', ...defaultTheme.fontFamily.sans],
      },
    }
  }
}
```

So if I understand correctly, this should be a typo?

### DIFF
--- a/src/pages/docs/font-family.mdx
+++ b/src/pages/docs/font-family.mdx
@@ -123,7 +123,7 @@ For convenience, [Preflight](/docs/preflight) sets the font family on the `html`
     theme: {
       extend: {
         fontFamily: {
-+         'sans': ['Proxima Nova', ...defaultTheme.fontFamily.sans],
++         'sans': ['"Proxima Nova"', ...defaultTheme.fontFamily.sans],
         },
       }
     }


jacobthesheep:
It's actually fine written like this

```
fontFamily: {
   sans: ['Rubik', ...defaultTheme.fontFamily.sans]
}
```

xuatz:
> It's actually fine written like this
> 
> ```
> fontFamily: {
>    sans: ['Rubik', ...defaultTheme.fontFamily.sans]
> }
> ```

But `Rubik` does not have a space like `'"Exo 2"'`, does it not matter? In that case should this section of the docs be removed?

https://tailwindcss.com/docs/font-family
image

adamwathan:
You only need the extra quotes when the font name has a character in it that isn’t valid in an identifier (like a number like in “Exo 2”) 👍🏻 

https://mathiasbynens.be/notes/unquoted-font-family

xuatz:
Got it, thanks for providing that link 👍

PR #1312, opened 2022-07-27
match comment color to demo color

SankThomas:
The comments explained a different color than was used in the demonstration. I've changed the comment to say the correct color.


bradlc:
Thanks!

PR #1313, opened 2022-07-31
feat: add copy to clipboard for selectors

zetske:
As a dev who uses Tailwind every day and who keeps the docs open 24/7 when working, this feature has been on my mind since day one. I find myself clicking on the selectors as a habit even though I realise it won't do anything. I'm sure I'm not the only one. 

My suggestion is to add the already implemented copy functionality to the classes components too. 

This PR is only a friendly suggestion and no feelings will get hurt if it gets thrown out. I'm open to suggestions and comments for improving on this PR.

Also, thank you for what you're doing. I'm a big fan of Tailwind.

https://user-images.githubusercontent.com/50806665/182046020-43f98303-bcab-4459-8027-cc3797bc1422.mov

### DIFF
--- a/src/components/ClassTable.js
+++ b/src/components/ClassTable.js
@@ -3,6 +3,7 @@ import { isObject } from '@/utils/isObject'
 import { castArray } from '@/utils/castArray'
 import clsx from 'clsx'
 import { Heading } from '@/components/Heading'
+import { CopyButton } from '@/components/Steps'
 
 function renderProperties(
   properties,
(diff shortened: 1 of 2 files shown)


thecrypticace:
Hey, definitely love the spirit of the idea but not sure about the implementation and little details. For instance, the top copy tooltip gets clipped completely. And the extra copy button doesn't feel like how we'd do it. But, in all honesty, I'm also not sure what we'd prefer here. An idea I thought about for a bit was a more implicit click-to-copy behavior for the selector itself but we still need to worry about specific design and functionality to give feedback to the user. I'm gonna close this because it's not really what we want to ship as-is. At the moment, neither I nor the rest of the team have the time to devote to this particular change (though we do like the idea) since we'd need to iterate on it and come up with something we feel comfortable shipping. Having said that, it is entirely possible that we revisit this in the future!

I really, really appreciate your contribution here and will keep this in mind if/when it comes up internally.

PR #1316, opened 2022-08-03
added section to fix #1315

### LINKED ISSUE #1315: Remix Tailwind integration missing a step
The TailwindCSS docs are missing this code section that the Remix docs include causing Tailwind to not work when using the Remix setup.
```js
import { Links } from "@remix-run/react";
// ...
export default function Root() {
  return (
    <html>
      <head>
        <Links />
        {/* ... */}
      </head>
      {/* ... */}
    </html>
  );
}
```

[Tailwind Docs](https://tailwindcss.com/docs/guides/remix)
[Remix Docs](https://remix.run/docs/en/v1/guides/styling)

brittneypostma:
This section will fix #1315

### DIFF
--- a/src/pages/docs/guides/remix.js
+++ b/src/pages/docs/guides/remix.js
@@ -107,6 +107,32 @@ export function links() {
     },
   },
   {
+    title: 'Import the Links component to render the links',
+    body: () => (
+      <p>
+        Import the Links component in your{' '}
+        <code>./app/root.jsx</code> file and use it inside the head section.
+      </p>
+    ),
+    code: {
+      name: 'root.jsx',
+      lang: 'js',
+      code: `import { Links } from "@remix-run/react";
+      // ...
+      export default function Root() {
+        return (
+          <html>
+            <head>
+              <Links />
+              {/* ... */}
+            </head>
+            {/* ... */}
+          </html>
+        );
+      }`,
+    },
+  },
+  {
     title: 'Start your build process',
     body: () => (
       <p>


thecrypticace:
Hey, thanks for the PR. The ultimate goal we had with the guides was to instruct people on how to set up a fresh project. We treat all guides this way otherwise we'd end up with lots of little extra notes and steps that are not necessary for the general case. With this goal in mind we've also tried to keep all the guides as similar as possible and only deviating where required. In Remix, the default structure imports and uses the `<Links />` component already. Given that, this addition isn't strictly necessary for the perspective of the guides that we're looking for. Gonna close this one but definitely appreciate the PR nonetheless. ✨

brittneypostma:
I mean I appreciate you being polite, but I have to disagree. I was using "Just the basics" when setting up the remix app and it did not have that configuration setup. I had to go over to the Remix TailwindCSS docs to actually get it working. Following the Tailwind docs produced no error, but was not working. I feel like this is a tradeoff of worse DX for consistency?

thecrypticace:
Can you tell me what settings you used? The `<Links>` element is present in `app/root.jsx` in the JS template and `app/root.tsx` in the typescript template (it's also in the production ready templates as well). When I reviewed this PR earlier I made sure to follow the guide and I also used "Just the basics" as you mention and it definitely does have them:

Screen Shot 2022-08-04 at 17 08 23
 
And it definitely works:

https://user-images.githubusercontent.com/614993/182953861-338eb47c-7ee6-48df-8696-ebfd93eca19f.mov

If it's missing when you install it I'd suggest running `create-remix` with the debug flag (`npx create-remix@latest --debug my-project`). It should should tell you where it's using the template from:

Screen Shot 2022-08-04 at 17 21 21

PR #1334, opened 2022-08-27
fix: SolidJS Example `className` attribute

ardeora:
Thanks for adding a SolidJS guide. The JSX example included in the last step uses the `className` attribute in the JSX. This is invalid SolidJS JSX syntax. We can just use `class` attribute directly here without any issue 😄

Screenshot 2022-08-27 at 1 15 51 PM


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Good catch!

PR #1335, opened 2022-08-28
Fix wrong file name in Nuxtjs guide

nicolasbeauvais:


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

But this file is already correct, notice the `build` section, that's the `build` section of your nuxt.config.js file not from your postcss.config.js file.

More info: https://nuxtjs.org/docs/configuration-glossary/configuration-build#postcss

image

PR #1336, opened 2022-08-30
Update content-configuration.mdx

rix1:
Add example showing how to make hot reloading work with Turborepo (symlinked packages)

See https://github.com/tailwindlabs/tailwindcss/discussions/7717#discussioncomment-2310984
for more details.

### DIFF
--- a/src/pages/docs/content-configuration.mdx
+++ b/src/pages/docs/content-configuration.mdx
@@ -275,6 +275,23 @@ If you're working in a monorepo with workspaces, you may need to use `require.re
   }
 ```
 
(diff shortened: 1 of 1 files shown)


rix1:
x-posting this from the discussion: 

> Hmm... After a bit of testing I think writing the path out in full (option 1) is better. Like, Tailwind CLI works perfectly, but Prettier (of all things) does not like this configuration. I guess the reason being that VScode is running Prettier from root, which means the UI package have to be installed at root as well 🥴 In addition, the Editor UI started lagging (e.g. search), probably because Prettier now have to import and resolve symlinks(?)

Closing this PR, as it's not a preferable solution. Absolute paths works better. This is my new `tailwind.config.js`

```js
content: [
    './pages/**/*.{ts,tsx}',
    './components/**/*.{ts,tsx}',
    '../../packages/ui/src/**/*.{ts,tsx}'
  ],
```

PR #1350, opened 2022-09-14
Update link to Next.js + Tailwind CSS Example

totti-rdz:
The previous link points to an old version of the Next.js + Tailwind CSS Example. With this old version it is not possible to simply one click copy the commands for each package manager (npm, yarn and pnpm), therefore I updated the link to the current (Sep 2022) version.


thecrypticace:
Thanks! ✨

PR #1354, opened 2022-09-20
Add default `font-weight` documentation to `font-size`

wongjn:
Seems like the `font-size` documentation is missing information about the [`font-weight` option](https://github.com/tailwindlabs/tailwindcss/blob/27aee8fc29c48b34068c546d5384802fc09aa0ed/src/corePlugins.js#L1636-L1644) – could we consider adding it in?

Alternatively, with the increasing number of properties, would it be worth consolidating the `line-height`, `letter-spacing` and `font-weight` sections somehow?

### DIFF
--- a/src/pages/docs/font-size.mdx
+++ b/src/pages/docs/font-size.mdx
@@ -77,34 +77,29 @@ Control the font size of an element using the `text-{size}` utilities.
 
 ### Customizing your theme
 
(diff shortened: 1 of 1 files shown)


reinink:
@wongjn Hey buddy, thanks for this contribution. I tweaked this PR a little, going with your alternative recommendation where we consolidate this all into a single section. I think this works pretty well 💪

Thanks again!

PR #1370, opened 2022-10-02
fix missing quotes for theme function

davidjbng:
Docs should use quotes within `theme` function.

Omitting the quotes still works, but the editor shows an error and there is no IntelliSense ([see Tailwind Play](https://play.tailwindcss.com/NKLx1ffCmG?file=css)).


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I'm going to close this for now since we actually do support that signature in Tailwind itself, but the intellisense extension is not updated to handle that yet.

Instead, I'm going to open an issue on intellisense, so that once it is fixed there, then it should be fixed on Tailwind Play and in your VSCode as well.

PR #1378, opened 2022-10-07
Add arbitrary-variants section in arbitrary-values

Sleepful:
When reading the section on `arbitrary variants` it is difficult to know that there are documents for `arbitrary variants`, which are very relevant. `arbitrary variants` are a little hidden, under a title that doesn't mention anything about CSS selectors, therefore this PR: Adds a link to the related content.

I added this right after `### Arbitrary properties` because it is almost as essential, and I added it before `### Handling whitespace` given that white space is a CSS selector and the order of question tends to be:
1. How do I add a custom selector?
2. How do I add the specific white space (descendant) selector?

### DIFF
--- a/src/pages/docs/adding-custom-styles.mdx
+++ b/src/pages/docs/adding-custom-styles.mdx
@@ -4,6 +4,7 @@ description: Best practices for adding your own custom styles to Tailwind.
 ---
 
 import { TipGood, TipBad, TipCompat, TipInfo } from '@/components/Tip'
+import { SnippetGroup } from '@/components/SnippetGroup'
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/adding-custom-styles.mdx
```diff
@@ -116,9 +116,13 @@ This can be useful for things like CSS variables as well, especially when they n
 </div>
 ```
 
+### Arbitrary variants
+
+In the same vein as arbitrary values, **arbitrary variants** allow you to write custom CSS selector modifiers, please refer to [arbitrary variants](/docs/hover-focus-and-other-states#using-arbitrary-variants).
```
Sleepful, line 121: I added the bold here because it's easy to gloss over the words given how similar they are to `arbitrary values`.

PR #1380, opened 2022-10-09
Adjust SVG icons' stroke color for the dark theme

itslouisgs:
The SVG icons blend with the background color when the theme is dark, as shown in the image below. To avoid this, I added `dark:stroke-white` class for those two SVG icons.
image

## Why `dark:stroke-white`?
As you can see from `ShowcaseLayout.js`, the title's `div` wrapper contains `text-slate-900 dark:text-white` classes. So, I assume that, in this case, `white` might be the best-suited dark theme variant of `slate-900`.


bradlc:
Thank you!

PR #1387, opened 2022-10-13
Update functions-and-directives.mdx

m7mark:
when using theme inside plugin, need to pass props in quotation marks


thecrypticace:
Not sure how we missed this 😅

Thanks!

PR #1411, opened 2022-10-26
clarify SvelteKit example

benmccann:
https://tailwindcss.com/docs/guides/sveltekit showed just the lines to be added, but wasn't clear about that, which confused people: https://github.com/sveltejs/kit/issues/4382#issuecomment-1292370424. Showing the full file is probably clearer


reinink:
Hey thanks for this contribution — I agree it's more clear to include this entire file.

Note, I've made a tweak to the code block formatting 👍

PR #1413, opened 2022-10-31
Fix `laravel` guide @ blade integr.

xerc:
w/o active `vite` ; "Unable to locate file in Vite manifest: resources/css/app.css."

### DIFF
--- a/src/pages/docs/guides/laravel.js
+++ b/src/pages/docs/guides/laravel.js
@@ -112,7 +112,7 @@ let tabs = [
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    @vite('resources/css/app.css')
+    @vite('resources/js/app.js')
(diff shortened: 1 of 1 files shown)


reinink:
Hey! So I just tested our Laravel Vite guide (https://tailwindcss.com/docs/guides/laravel), and it worked perfectly following the steps outlined in that guide.

Switching from `@vite('resources/css/app.css')` to `@vite('resources/js/app.js')` actually breaks the site styles, as they are no longer loaded. This approach does work if you import your `app.css` file into your `app.js` file, like this:

```js
// ./resources/js/app.js
import "../css/app.css";
import "./bootstrap";
```

However, I think that's beyond the scope of this guide. We try to keep these guides as simple as possible, and we want people to be able to follow the steps exactly as they are in the guides and get correct results. That does however mean that you may have to adjust things for your own particular project and setup.

Hope that helps 👍

PR #1420, opened 2022-11-07
Fix 404 Editor Tools image

wongjn:
Fixes a 404 error for an image on the homepage:
image

## Before:
image

## After:
image

---

Perhaps we could also consider adding `loading="lazy"` so the image is not requested at all in dark mode (since it has `dark:hidden` on it anyway)?


bradlc:
Thanks @wongjn! ❤️ Will definitely consider adding `loading="lazy"` as well 👍

PR #1439, opened 2022-11-23
Add Click-to-Copy to ClassTable classes 

tycrimm:
A common flow is to open the docs, see the class you're looking for, copy it, and paste it into your editor of choice. This usually means a triple-click to select the text (or single click + dragging your cursor), followed by a quick `Control/Command + C`. Yes it works, but it may also result in silent tears of inefficiency. 😢 

What if there was a 1-click solution to it all?! What if we could turn 3 clicks (or 1 click + drag) and a keyboard combo into a single click! 😱 💥 🔥 

This PR breaks `CopyButton` out into its own component (from `Steps.js`) and adds it to `ClassTable`, enabling our one-click copying dream! 😍 

An open question is if this should be implemented via the `CopyButton` or if adding a custom button + hover states to the classnames as a "click to copy" variation would be cleaner from a UX perspective. Happy to make any changes, just want to save as many clicks as possible. 🖱️ Thanks!  

image

### DIFF
--- a/src/components/ClassTable.js
+++ b/src/components/ClassTable.js
@@ -3,6 +3,7 @@ import { isObject } from '@/utils/isObject'
 import { castArray } from '@/utils/castArray'
 import clsx from 'clsx'
 import { Heading } from '@/components/Heading'
+import { CopyButton } from '@/components/CopyButton'
 
 function renderProperties(
   properties,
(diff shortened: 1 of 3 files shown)


reinink:
@tycrimm Hey! Thanks for this contribution.

While I see the value in adding this kind of functionality to the website, I think we'd want to look at some alternative approaches to the design, which unfortunately isn't something we have time for at this particular moment, so I think I'm going to close this PR for now until we have time to look at this more closely.

Thanks either way for your efforts! 🙏

PR #1452, opened 2022-12-03
fix(index): use Title component

KayBeSee:
refactors src/pages/index.js to use Title component from Meta


adamwathan:
Hey! This doesn't seem like it works, as per the preview deploy generated by the PR:

image

So going to close this. Currently on the live site the title and OG title information is all correct for this page, so nothing to actually fix.

PR #1455, opened 2022-12-09
Update SvelteKit guide for 1.0

benmccann:
We will soon be releasing SvelteKit 1.0. The `latest` tag will work now with the latest betas / RCs and will continue to work after 1.0. The `next` tag will only work until 1.0 is released


thecrypticace:
Confirmed it works as expected. Thanks! ✨

PR #1461, opened 2022-12-12
Update guide for appDir feature

thiagodebastos:
Following the instructions does not work if using NextJS's experimental `appDir` feature.


bradlc:
Hey! We actually have a whole separate version of the guide for people who want to use the `./app` directory here:

https://tailwindcss.com/docs/guides/nextjs#app-directory

image

Going to stick with that approach but thanks for the contribution regardless 👍🏻

PR #1464, opened 2022-12-16
The missed comma in the code example

AliRoshaniDev:
In these code examples, a comma is forgotten in the line that causes an error when executed.


reinink:
Hey good catch, thanks! 👍

PR #1476, opened 2023-01-04
Add callout to potentially previous mix alias

w0rd-driven:
In troubleshooting walking through the guide with a friend who is new to Phoenix, they had errantly added `"assets.deploy": ["tailwind default --minify", "esbuild default --minify", "phx.digest"]` below the existing alias. This had the effect of not building the expected `app.css` even though everything else was configured correctly.

I'm not a huge fan of how splitting this into 2 paragraphs breaks existing convention but I don't know a better way of separating the text or displaying it in a way that conveys the same idea. I figured if this tripped up one newbie it's bound to be a problem for a few others.

### DIFF
--- a/src/pages/docs/guides/phoenix.js
+++ b/src/pages/docs/guides/phoenix.js
@@ -61,11 +61,9 @@ let steps = [
   {
     title: 'Update your deployment script',
     body: () => (
-      <div>
-        <p>
-          Configure your <code>assets.deploy</code> alias to build your CSS on deployment.
-        </p>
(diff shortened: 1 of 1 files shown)


w0rd-driven (commented):
That's a much better approach 🙏. You may want to revert that parent div in the body since there's no longer a reason to nest the now singular paragraph tag.

thecrypticace:
Hah yep good call

thecrypticace:
Thanks!

PR #1509, opened 2023-02-15
Use 4-space indent in Laravel examples

datlechin:
In the `.editorconfig` config file, Laravel uses `indent_size = 4` for all file types (js, html, ...). So this PR will use 4-space indentation for Laravel's Tailwind CSS implementation examples.


reinink:
Hey! Thanks for this suggestion — I've actually wondered about making this change myself in the past. However, I think we're going to keep it the way it is, just because all the other framework guides all use 2-spaces, plus it allows you to see more of the code without scrolling.

Assuming you have something like Prettier setup in your Laravel project, it will auto format to 4-spaces on save either way.

Thanks for the suggestion regardless! 👍

PR #1514, opened 2023-02-18
Add section for TypeScript monorepo configuration

zachkirsch:
Hey! New user of Tailwind and already it's so much better than the css I was writing before. I have probably written these lines thousands of times:

```css
.container {
  display: flex;
  flex-direction: column;
}
```

Anyway, when trying to set up tailwind in our vite+typescript monorepo, it immediately broke HMR (every change caused a full reload). I was able to fix this by change `content` in `tailwind.config.js` to point to our source TS files, instead of the generated JS files.

This PR includes an update to the docs, to save other people the effort to figure this out.

### DIFF
--- a/src/pages/docs/content-configuration.mdx
+++ b/src/pages/docs/content-configuration.mdx
@@ -331,6 +331,21 @@ If you're working in a monorepo with workspaces, you may need to use `require.re
   }
 ```
 
+If you're working in a _TypeScript_ monorepo with workspaces, you may need to point to your source files for Hot Module Replacement to work.
+
+```js tailwind.config.js
+  const path = require('path');
+
+  module.exports = {
+    content: [
+      './components/**/*.{html,js}',
+      './pages/**/*.{html,js}',
+>     path.join(path.dirname(require.resolve('@my-company/tailwind-components')), '../src/**/*.{ts,tsx}'),
+    ],
+    // ...
+  }
+```
+
 ### Using relative paths
 
 By default Tailwind resolves non-absolute content paths relative to the current working directory, not the `tailwind.config.js` file. This can lead to unexpected results if you run Tailwind from a different directory.


w3bdesign (approved):

thecrypticace:
@zachkirsch I'm not entirely sure how this is much different than the section above it. Do you have an example project demonstrating the situation where this becomes important?

zachkirsch:
@thecrypticace [here's](https://github.com/fern-api/fern/blob/e5cd749267138655bcdb44a1796468471534ad9e/packages/ui/fe-bundle/tailwind.config.cjs#L8) an example.

>  I'm not entirely sure how this is much different than the section above it

The section above it recommends doing:

```js
path.join(path.dirname(require.resolve('@my-company/tailwind-components')), '**/*.js')
```

`require.resolve` is a runtime API to determine where the javascript files lives. In a JS monorepo, this is `<path to pacage>/src`, but for a typescript monorepo, this is something like `<path to package>/lib` (TS files live in `src/` and are compiled into `lib/`). It still _works_ (i.e. the styles are applied), but now that we're watching the compiled JS files instead of the source files, it screws up HMR. Every file change becomes a full page reload.

IMO it's a non-obvious fix which is why I added this PR!

reinink:
Hey @zachkirsch, thanks for sharing this PR. We've been hesitating on merging it in because on one hand it's helpful information for people in the same situation as you to know, but on the other hand it's a rather specific scenario that I don't think that many people are going to run into.

I think in the interest of keeping the documentation simple we're going to leave this out for now, but often just having a PR like this around is helpful for folks who bump into this and do a search. And if we do get a bunch of reports of others running into the same problem, we can always revisit this decision.

Hope that makes sense 👍

zachkirsch:
100 percent!

PR #1517, opened 2023-02-20
Highlight content paths

abdul-alhasany:
This is an important information that users need to pay attention to. It might be missed when a user reading the page.


adamwathan:
Hey! I don't think this change makes sense since this content isn't a blockquote and shouldn't be styled like one:

image

Thanks though.

abdul-alhasany:
I looked up markdown alert or notice syntax in but could not find anything except this. I think any style that brings attention to it is fine. I think it is a vital piece of information.

PR #1528, opened 2023-02-28
fix: add the Tailwind directives to main.css instead of style.css

ehsan-shv:
Based on the most recent Vite release, it creates an assets directory in src in Vue apps and requires the import of the Tailwind directives in src/assets/main.css.

### DIFF
--- a/src/pages/docs/guides/vite.js
+++ b/src/pages/docs/guides/vite.js
@@ -173,11 +173,11 @@ let tabs = [
         body: () => (
           <p>
             Add the <code>@tailwind</code> directives for each of Tailwind’s layers to your{' '}
-            <code>./src/style.css</code> file.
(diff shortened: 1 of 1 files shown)


reinink:
Hey! I just gave this a try and it appears that the `src/style.css` file still exists when you do a fresh Vite install using the Vue template:

```
npm create vite@latest my-project -- --template vue
```

image

So I think it's best to leave the guide as it is, but if this ever does change we will update it to match then.

Thanks either way! 👍

ehsan-shv:
Run `npm create vite@latest` and select "Customize with create-vue":

reinink:
@ehsan-shv Ahh yes, if you choose a different template then the installation instructions will need to be tweaked. This is why we've explicitly chosen the `react` and `vue` templates for the Vite framework guide 👍

PR #1537, opened 2023-03-04
Documents logical properties [WIP]

danhollick:
Adds examples for:
 - [x] Top / Right / Bottom / Left
 - [x] Margin 
 - [x] Padding 
 - [x] Border Radius
 - [x] Border Width
 - [x] Border Color
 - [x] Scroll Padding
 - [x] Scroll Margin
 
 I'm still working on the best way to structure the component.


danhollick:
So the content is finished here but I'd love some help with the component architecture. 

So the way I've set it up at the moment is a main `<LogicalProperties/>` component that accepts a `property` prop and switches the internal component based on that:


```jsx
/* .mdx */
<LogicalPropertyExample property="padding"/>
...

 ```
 
 ```jsx
 /* LogicalProperty/index.js*/
 ... 
 <ExampleWrapper direction={direction} handleDirectionChange={setDirection}>
        {property === 'margin' && <MarginExample direction={direction} />}
        {property === 'padding' && <PaddingExample direction={direction} />}
        ...
```

The issue is that currently the code snippet for the example is still in the `.mdx` file:

```jsx
<LogicalPropertyExample property="padding"/>

```html
<div class="flex ...">
  <div class="**ps-4** ...">ps-4</div>
  <div class="**pe-12** ...">pe-12</div>
</div>
``
```
I'd prefer it was co-located with the example component but not sure how to do it. Had a look at some of the other components and I'm not sure I know enough about remark/rehype to know what's going on there.

danhollick:
One other thing:

> Currently, this is only supported in the inline direction

This doesn't feel true for Border Radius because you can specify a corner using both axes.

Screenshot 2023-03-06 at 09 50 26

PR #1554, opened 2023-03-23
Improve DX installation

matthieumota:
Hi,

I think it can be a good idea to add vendor path on content for Laravel pagination and Symfony theme form. Maybe it's usefull for other frameworks but I don't use it.

It is developer responsability to add the path for sure and it's not sure everybody use pagination / form in that projects but it's okay for all the use case.

### DIFF
--- a/src/pages/docs/guides/laravel.js
+++ b/src/pages/docs/guides/laravel.js
@@ -59,6 +59,7 @@ let tabs = [
 >     "./resources/**/*.blade.php",
 >     "./resources/**/*.js",
 >     "./resources/**/*.vue",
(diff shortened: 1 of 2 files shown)


reinink:
Hey! Thanks for this suggestion, but I think we're going to keep this guide the way it is for now, just because our goal with these guides is to provide a very simple minimal set of instructions on how to get up and running with a particular framework, and they are not really designed to cover all the possible situations someone might run into. If we did that you could imagine how quickly these guides could become very long and complex, and difficult to maintain. Hope that makes sense 👍

PR #1556, opened 2023-03-24
Update Footer.js links to pop open in a new tab.

Bowens20832:
Updated the footer navigation links to pop open external websites in a new tab. This helps users with not losing their place on Tailwindcss.com, while they can explore the other Tailwind resources in another browser window.

I also updated the external urls to use the 'noreferrer' tag so that Tailwind isn't passing valuable SEO page rank to GitHub and Twitter.

### DIFF
--- a/src/components/home/Footer.js
+++ b/src/components/home/Footer.js
@@ -34,8 +34,8 @@ export function Footer() {
                   <ul className="mt-3 space-y-2">
                     {items.map((item) => (
                       <li key={item.href}>
-                        <Link href={item.href}>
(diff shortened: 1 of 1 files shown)


RobinMalfait:

I'm going to close it for now because this also means that when users _do_ want to open the link in the current page that they can't anymore, however the inverse behavior is still possible.

You can open links in a new tab already if you hold cmd/ctrl and click or if you use the scroll wheel on your mouse (aka middleclick).

With the current behaviour we can satisfy both types of users.

Hope this makes sense!

PR #1559, opened 2023-03-28
Remove template string interpolation from example

ryanstaniforth:
Small fix for some missing backticks on the example


thecrypticace:
Oops. Didn't mean to write it with the interpolation stuff. Good catch. I also fixed the indentation.

PR #1560, opened 2023-03-29
Update emberjs.js

allthesignals:
Today, I was using Tailwind to try something out, and this code broke my build:

```handlebars
<div class="bg-[url('/sample.png')]"></div>
```

Not knowing much about Webpack, I was surprised to see it even trying to do anything with `sample.png`:
```
webpack 5.76.3 compiled with 1 error in 5311 ms
Build Error (PackagerRunner) in sample.png

Module parse failed: Unexpected character '�' (1:0)
```

After some time I learned I needed to tell Webpack how to process PNGs. Seems reasonable!

I think it should be mentioned during installation as this current setup won't work with all of CSS/Tailwind's features.

### DIFF
--- a/src/pages/docs/guides/emberjs.js
+++ b/src/pages/docs/guides/emberjs.js
@@ -79,6 +79,17 @@ let steps = [
 >                 },
 >               ],
 >             },
+>            {
+>              test: /\.(jpg|jpeg|png|woff|woff2|eot|ttf|svg)$/,
+>              use: [
+>                {
+>                  loader: 'url-loader',
+>                  options: {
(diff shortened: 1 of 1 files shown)


reinink:
Hey! Thanks for this suggestion, but I think we're going to keep this guide the way it is for now, just because our goal with these guides is to provide a very simple minimal set of instructions on how to get up and running with a particular framework, and they are not really designed to cover all the possible situations someone might run into. If we did that you could imagine how quickly these guides could become very long and complex, and difficult to maintain. Hope that makes sense! 🙏

PR #1564, opened 2023-03-31
Fix typo in v3-3 announcement

irajsuhail:
In the logical properties section, the Physical counterpart (LTR) of `rounded-e-*` is `rounded-r-*` rather than `rounded-l-*`


RobinMalfait:
Hey good catch, thanks so much for this contribution! 🙏

PR #1577, opened 2023-04-11
Added query params to search the docs

Fabian9799:
Added the option to search the docs using `?q=width`
image

I usually make my projects in svelte so my code is probably not optimal and can be improved but it works

### DIFF
--- a/src/components/Search.js
+++ b/src/components/Search.js
@@ -18,6 +18,17 @@ export function SearchProvider({ children }) {
   const [isOpen, setIsOpen] = useState(false)
   const [initialQuery, setInitialQuery] = useState(null)
 
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
+      const searchParams = new URLSearchParams(window.location.search);
+      const query = searchParams.get('q');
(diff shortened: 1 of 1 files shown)


reinink:
Hey! Thanks for this contribution, but I generally think it's more useful to just link folks directly to a specific page (or multiple pages) in the docs rather than to a preset search. So in the interest of just keeping this project as simple as possible, I think we're going to pass on this addition for now. Hope that makes sense 🤙

Fabian9799:
I totally get it that linking to a page directly is better.

My addition was meant for individual users since most browsers now support [Site Search](https://support.google.com/chrome/answer/95426) which allows someone to search directly in the url bar without needing to go to a website and then clicking on search

Here is an example with youtube: 
image
which redirects me to https://www.youtube.com/results?search_query=tailwind

That way I don't need to go to youtube and click the searchbar as I can directly type `youtube.com` if I have it configured inside my browser settings

This article explains it a bit better then the official chrome docs: https://chromeunboxed.com/chrome-site-search-shortcuts

PR #1579, opened 2023-04-15
Fix theme update

microdev1:
In the theme `update` function, instead of adding `changing-theme` class, it was being removed when going from `dark` to `light`.


bradlc:
Thank you!

PR #1586, opened 2023-04-27
Removed duplicate gap class

benfoster:
Spotted a couple of duplicate classes when I was working through some of the examples in the docs site.


bradlc:
Thanks!

PR #1587, opened 2023-04-27
Fix example code snippet

kleviss:
Close the image tag in the provided code snippet for the card.


RobinMalfait:

However going to close this PR because an `img` tag is a [`void` element](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) which means that it cannot have children and doesn't require a `/`. It currently already is valid HTML.

PR #1610, opened 2023-05-27
Implement custom id syntax on headings

nacal:
I'm now trying to create Japanese documentation by forking this repository. It is to let more people in Japan use my favorite Tailwinscss :)

In translating to other languages, I need a custom id attribute in Heading. It is to maintain existing links.
It is possible to achieve this only in my environment forked, but this is very useful, so I created a PR.

Usage
```
## Heading {/*heading-id*/}

=> <h2 id="heading-id" ...><a href="#heading-id" ...>Heading</a></h2>
```

This is also accomplished in the React documentation, and the implementation was based on that.
https://github.com/reactjs/react.dev/blob/main/plugins/remark-header-custom-ids.js

Please let me know if you can already achieve this in other ways.
Thanks for your maintenance.

### DIFF
--- a/remark/withTableOfContents.mjs
+++ b/remark/withTableOfContents.mjs
@@ -18,13 +18,18 @@ export const withTableOfContents = () => {
         )
         let slug = slugify(title)
 
+        if (node.children[node.children.length - 1].type === 'mdxTextExpression') {
+          slug = node.children.pop().value
+          slug = slug.slice(2, slug.length - 2)
+        }
+
         let allOtherSlugs = contents.flatMap((entry) => [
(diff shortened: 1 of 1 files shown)


bradlc:
Hey @nacal, this is cool but I'm going to say no on this occasion just because we don't have a need for it within the `tailwindcss.com` project right now. I hope that's understandable!

For what it's worth, I think if we did need to define custom heading IDs we would use something like [`mdx-annoations`](https://github.com/bradlc/mdx-annotations) which would allow us to specify them like this:

```
# Hello world {{ id: 'intro' }}
```

PR #1617, opened 2023-06-13
Adds RedwoodJS to Framework Guides

cannikin:
We've got first-class Tailwind support and thought we'd make a great addition to the Installation > Framework Guides 🙂


cannikin:
Hey @reinink thanks for the feedback. I copied the [Astro guide](https://tailwindcss.com/docs/guides/astro) so I thought a one-line installer was fine! We do require `yarn` for our stuff.

dthyresson:
While I understand this stance, I'd argue that if the goal is to show how best to use TailwindCSS with each framework, then it's best to showcase its integration in the best and most seamless way. Personally, I don't need to know how to setup TailwindsCSS (and if I want to know the inner details, I can read the standard Node docs) -- I just want to get started using this amazing tool fast and w/out mistakes and the tedium of config and package install (and support issues/request because I had a config wrong or missed a step).

This is what Astro's docs do: https://tailwindcss.com/docs/guides/astro

image

and the details: https://github.com/withastro/astro/blob/main/packages/integrations/tailwind/src/index.ts

As for yarn vs npm, RedwoodJS favors yarn over npm due to its need for workspaces. Again, document the Framework guide the best way for the framework as appropriate.  Obviously the Rails guide doesn't use npm because it isn't JS, but it uses what is best and appropriate for its guide.

reinink:
> I copied the [Astro guide](https://tailwindcss.com/docs/guides/astro) so I thought a one-line installer was fine! We do require yarn for our stuff.

> This is what Astro's docs do: https://tailwindcss.com/docs/guides/astro

Hey, totally understand the confusion with the Astro framework guide. To be honest, we would have preferred to stick with our normal steps for the Astro guide as well, however, in the case of Astro that was unfortunately not possible due to how their SSR pipeline works. The Astro plugin does meaningful setup for SSR that can't be done without their plugin. If that wasn't necessary, we would have shown our normal steps — we only deviate when absolutely necessary.

> While I understand this stance, I'd argue that if the goal is to show how best to use TailwindCSS with each framework, then it's best to showcase its integration in the best and most seamless way. Personally, I don't need to know how to setup TailwindsCSS (and if I want to know the inner details, I can read the standard Node docs) -- I just want to get started using this amazing tool fast and w/out mistakes and the tedium of config and package install (and support issues/request because I had a config wrong or missed a step).

Heh, yup, understand where you're coming from here too. Part of this also comes down to the fact that we don't have control over third-party tools and how these quick installers work. Sometimes they're great, but sometimes they're not so great and actually limit Tailwind CSS, or have issues that we as the maintainers of the project end up having to help with, but since we're not familiar with all these tools it's harder for us to help.

In the end we're better able to help folks if everyone more-or-less uses the same installation story for Tailwind CSS.

The other benefit here is that the Tailwind CSS installation story is also (reasonably) the same going from one framework to the next, making this knowledge more transferable.

Totally get that this is a tradeoff though 👍

> As for yarn vs npm, RedwoodJS favors yarn over npm due to its need for workspaces. Again, document the Framework guide the best way for the framework as appropriate. Obviously the Rails guide doesn't use npm because it isn't JS, but it uses what is best and appropriate for its guide.

Understood, if Yarn is necessary for RedwoodJS, then it's fine for the guide to use it.

PR #1623, opened 2023-06-23
Homepage "Dark Mode" section grammar fixes

wackbyte:


thecrypticace:
Hey! Appreciate the PR but we're going to leave this as is for now.

Our use of "then" is grammatically correct as it is being used as [an indicator for "next"](https://dictionary.cambridge.org/us/grammar/british-grammar/then). And the time `2am` being lowercase is purely a stylistic choice.

But appreciate the PR nonetheless — thanks!

PR #1626, opened 2023-06-26
add arbitrary example for bg-position

jabranr:
I spent too long figuring out how to set/use the arbitrary values for `background-position` for background, especially helpful when trying to set this for multiple backgrounds. So I thought it should make an excellent addition to the docs. Likely to be helpful for others landing here with similar issues.


jabranr:
That was in the wrong file. I will create a new PR with correct change

PR #1627, opened 2023-06-26
Add 'select-all' to ClassTable

tycrimm:
This PR adds the `select-all` class to the class names displayed in the ClassTable component.

Why?

#1439 was closed for future design consideration in November.  I'm sadly still triple-clicking to select classes I want to copy in the docs. 

This PR reduces it to a single click (a savings of 66%!) before the mighty `Command + C` is invoked.

Also happy to implement this as "DoubleClick => Class Copied to Clipboard" but figuring this was a safer initial approach. Please let me know!

### DIFF
--- a/src/components/ClassTable.js
+++ b/src/components/ClassTable.js
@@ -132,7 +132,7 @@ export const ClassTable = memo(
                         <td
                           translate="no"
                           className={clsx(
-                            'py-2 pr-2 font-mono font-medium text-xs leading-6 text-sky-500 whitespace-nowrap dark:text-sky-400',
+                            'py-2 pr-2 font-mono font-medium text-xs leading-6 text-sky-500 whitespace-nowrap dark:text-sky-400 select-all',
                             {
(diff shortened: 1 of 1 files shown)


adamwathan:
Hey! Thanks for the suggestion but going to say no for now — personally it just feels unexpected to me that when I go to click in that area that the text is automatically selected.

If this becomes a super common complaint we can look into prioritizing coming up with a solution we are happy with 👍

tycrimm:
@adamwathan Totally agree that this is unexpected. Would you be open to adding a "Click to Copy" hover interaction similar to #1439 ? I'm happy to tee everything up/work with whoever.. just want to avoid making a Chrome Extension to add this myself 😂 

image

Rough Math: [SimilarWeb estimates](https://www.similarweb.com/website/tailwindcss.com/#overview) tailwindcss.com received 5.6M visits this past month. If 1% of those visits result in someone copying a class, that's 1.12M clicks that can we saved for Flappy Bird. I have a vested interest in saving what seems like 20,000 of those clicks 🤦‍♂️

PR #1628, opened 2023-06-26
add info about custom arbitrary usage

jabranr:
I spent too long figuring out how to set/use the arbitrary values for background-position for background, especially helpful when trying to set this for multiple backgrounds. So I thought it should make an excellent addition to the docs. Likely to be helpful for others landing here with similar issues.

### DIFF
--- a/src/pages/docs/background-position.mdx
+++ b/src/pages/docs/background-position.mdx
@@ -155,3 +155,9 @@ Learn more about customizing the default theme in the [theme customization](/doc
 ### Arbitrary values
 
 <ArbitraryValues property="background-position" featuredClass="bg-[center_top_1rem]" />
+
+If you need further control over background position then you can use `bg-[position:...]` to customize the outcome. It is also useful to adjust positions of multiple backgrounds. i.e.
+
+```html
+<div class="bg-[position:right_100px_bottom_-2rem]"></div>
+```


reinink:
Hey! Thanks for this contribution, and sorry to hear it took you a bit to figure this out.

I don't think we can merge this in as is, as this only updates the arbitrary values section of the background position documentation page. If we were to draw more attention to the arbitrary value types, I think it would make sense to do this for all utilities where this is relevant. However, to be honest, I'm not sure that we need to go that far.

Right now each arbitrary value section already links to the extended [arbitrary values](https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values) documentation, which includes a section on [resolving ambiguities](https://tailwindcss.com/docs/adding-custom-styles#resolving-ambiguities). This explains how you can use types to tell Tailwind specifically what property you're targeting.

These types are based on CSS data types, which you can find listed on MDN here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types

I think it would be helpful to some day list these types in the Tailwind documentation — even just the common ones.

I totally get how it would have been helpful to have more information about arbitrary value types directly on the background position page when you were running into this issue. However, as documentation maintainers it's hard to know how much information to include on each page versus when to just link elsewhere to learn more. For example, if were were to cover the arbitrary value types on each utility page, should we also explain how to use whitespace in them as well? It's a bit of a balancing act...

Hope that all makes sense 🤙

PR #1647, opened 2023-07-25
Improve heading anchor links

### LINKED ISSUE #1644: [BUG] Copy tag on page
**What browser are you using?**

Chrome 114.0.5735.248

**What operating system are you using?**

Windows 11 Pro

**Reproduction URL**

https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-peers

But it's happened on each page.

**Describe your issue**

When i hover on label, then tag is displayed but on the scroll from navigation aside.

image

it happened on width 1024px to 1180px and width >= 1280px.

reinink:
Resolves #1644

This PR improves the heading anchor links.

## Fixed anchor icon interfering with scrollbar

First, I tweaked the position of the sidebar navigation scrollbar so it's more to the left giving a little more space for the anchor link icon, and also moved the icon slightly to the right, so that the icon doesn't land on top of the scrollbar:

| Before | After |
|--------|--------|
| image | image | 

## Removed anchor icon on mobile

Next, I removed the anchor link icon entirely on mobile (since there is no space for it), but then made the whole heading a link so that you can still click it.

| Before | After |
|--------|--------|
| image | image |

## Changed hit/hover area

I also updated the hit/hover area to just the text, not the full width of the heading. To me it feels weird when I'm hovering over the heading on the right-hand side of the page, and the anchor link is appearing on the left.

image

## Removed old props

Finally I removed the `number` and `badge` props from the `<Heading>` component — pretty sure these aren't used anywhere and are remnants of the previous version of the website.

### DIFF
--- a/src/components/Heading.js
+++ b/src/components/Heading.js
@@ -35,11 +35,8 @@ export function Heading({
       data-docsearch-ignore={ignore ? '' : undefined}
       {...props}
     >
-      <a
-        className={clsx('group relative', hidden ? 'sr-only' : '-ml-2 pl-2')}
-        href={`#${id}`}
-      >
         <div className="absolute -ml-8 hidden items-center border-0 opacity-0 group-hover:opacity-100 lg:flex">
+      <a className={clsx('group relative', hidden ? 'sr-only' : '-ml-2 pl-2')} href={`#${id}`}>
           &#8203;
           <div className="flex h-6 w-6 items-center justify-center rounded-md text-slate-400 shadow-sm ring-1 ring-slate-900/5 hover:text-slate-700 hover:shadow hover:ring-slate-900/10 dark:bg-slate-700 dark:text-slate-300 dark:shadow-none dark:ring-0">
             <svg width="12" height="12" fill="none" aria-hidden="true">
--- a/src/components/Heading.js
+++ b/src/components/Heading.js
@@ -35,7 +35,10 @@ export function Heading({
       data-docsearch-ignore={ignore ? '' : undefined}
       {...props}
     >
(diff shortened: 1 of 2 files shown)


src/components/Heading.js
```diff
@@ -38,14 +35,14 @@ export function Heading({
...
-      {!hidden && (
-        <a
-          href={`#${id}`}
-          className="absolute -ml-10 flex items-center opacity-0 border-0 group-hover:opacity-100"
-          aria-label="Anchor"
-        >
+      <a
+        className={clsx('group relative', hidden ? 'sr-only' : '-ml-2 pl-2')}
+        href={`#${id}`}
+        aria-label="Anchor"
```
bradlc, line 41:
```suggestion
```

We need to remove this so that the heading text itself becomes the label

bradlc:
Looking good! One minor thing, I think the focus outline looks a bit odd:

CleanShot 2023-07-25 at 19 53 23@2x

I'm thinking the simplest solution could be to just make sure the icon thing appears on focus:

CleanShot 2023-07-25 at 19 54 49@2x

reinink:
> Looking good! One minor thing, I think the focus outline looks a bit odd:
> 
> CleanShot 2023-07-25 at 19 53 23@2x
> I'm thinking the simplest solution could be to just make sure the icon thing appears on focus:
> 
> CleanShot 2023-07-25 at 19 54 49@2x

Good call, fixed 👍

bradlc:
LGTM!

PatrykKuniczak:
I'm really impressed.

I see in this repo is several "legacy," issues, but it's a really marginal, this fix is made immediately, that's the best open source project, which i've seen, also the docs is very good.

😁

PR #1651, opened 2023-07-27
Add example extending min-h and min-w to use the spacing scale

raimondlume:
Quite often I find myself wanting to use the default spacing scale for `min-h` and `min-w`, i.e. `min-h-16` and have wondered multiple times why these aren't added by default.

I've always assumed there's a good reason behind it, but decided to look into it and found this comment by Adam: https://github.com/tailwindlabs/tailwindcss/discussions/7600#discussioncomment-5356444

From what I understand, the only reason was performance / bundle size pre-JIT.

I'd like to assume that having a min-height or width is something that isn't too rare, so I'd propose mentioning this in the docs under the [Using custom values](https://tailwindcss.com/docs/min-width#using-custom-values) section as this is can be a common point of confusion

wdyt?

image

### DIFF
--- a/src/pages/docs/min-height.mdx
+++ b/src/pages/docs/min-height.mdx
@@ -62,7 +62,7 @@ Learn more about customizing the default theme in the [theme customization](/doc
 
 ### Using the default spacing scale
 
-The default spacing scale (i.e. `min-h-16`) is not included by default, but by defining the option as a function you can access the theme which lets you target other properties of the theme, 'spacing' in this case.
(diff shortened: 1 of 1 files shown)


RobinMalfait:

I am going to close it for now because I think what we would rather do is revisit the default values for the next version of Tailwind CSS. We have an internal note for this 👍

We typically don't have this kind of explanation for each utility, however we do have a more general page where we explain how you can reference other values: https://tailwindcss.com/docs/theme#referencing-other-values

Thanks again!

PR #1653, opened 2023-08-01
Fix unexpected installations tabs scroll

gilhanan:
Fix unexpected installations tabs scroll which is caused by the negative margin on the anchor menu items.

Current:
image

After:
image


RobinMalfait:

While this PR gets rid of the scrollbar it also changes the styles of the active tab.

image
image

Notice how the active line is now thin and sitting above the gray line instead of on top of it.

---

I didn't have push rights on this branch so I created a new PR and started from this PR as a base to ensure that you are still a contributor. In the other PR (#1654) I ensured that the styles are the same as before but the scrollbars are gone.

Thanks again!

PR #1663, opened 2023-08-22
fix: added example to theme extend paragraph

DavideFrancescon:
Closes: #1571 

Result in the docs:

Screenshot 2023-08-22 alle 16 32 01

Result of the code in the example:

Screenshot 2023-08-22 alle 16 30 26

PR #1669, opened 2023-09-01
Add explanation of `theme.extend` to theme documentation

### LINKED ISSUE #1571: Tailwindcss documentation does not explain how to access properties defined under theme.extends
Let's consider a [grid-template-columns](https://tailwindcss.com/docs/grid-template-columns) example as follows:
```js
// tailwind.config.js
module.exports = {
  theme: {
    extend: {
      gridTemplateColumns: {
        // Simple 16 column grid
        '16': 'repeat(16, minmax(0, 1fr))',

        // Complex site-specific column configuration
        'footer': '200px minmax(900px, 1fr) 100px',
      }
    }
  }
}
```

**LET ME ASK YOU** : Is it so difficult to give a simple example aside as you go by reading the docs over, rather than googling, stackoverflowing reading some people examples out of between-the-lines, that accessing properties defined within `theme.extend.defaultKey.arbitrary_name` alike signature could be achieved just by doing the following i.e.:

```html
<div class="grid grid-cols-16">
    <!--  -->
</div>

<div class="grid grid-cols-footer">
    <!--  -->
</div>
```

The only implicitly mentioned thing is:
> Use the grid-cols-{n} utilities to create grids with n equally sized columns
> OP: _at least then say that -{n} could be also used as reference to some arbitrary-value defined under theme.extend.defaultKey_

[List of defaultKey (Default Keys)](https://tailwindcss.com/docs/theme#configuration-reference)

The following does not contain information about what I've just mentioned above . 
> Learn more about customizing the default theme in the [theme customization](https://tailwindcss.com/docs/theme#customizing-the-default-theme) documentation.

thecrypticace:
We talk about how to extend the theme but never discuss how you use the values within it. This PR adds a short snippet to the bottom of the docs discussing this.

Fixes #1571
Supersedes #1663

### DIFF
--- a/src/pages/docs/theme.mdx
+++ b/src/pages/docs/theme.mdx
@@ -222,7 +222,7 @@ module.exports = {
 }
 ```
 
-After adding this to your theme you can use it just as with any other utility in your HTML:
+After adding this to your theme you can use it in your HTML just like any other utility:
 
 ```html
 <h1 class="**font-display**">
--- a/src/pages/docs/theme.mdx
+++ b/src/pages/docs/theme.mdx
@@ -222,7 +222,7 @@ module.exports = {
 }
 ```
 
-After adding this to your theme you can use it in your HTML just like any other utility:
+After adding this to your theme you can use it just like any other `font-{family}` utility:
 
 ```html
 <h1 class="**font-display**">
--- a/src/pages/docs/theme.mdx
+++ b/src/pages/docs/theme.mdx
@@ -226,12 +226,12 @@ module.exports = {
 After adding these to your theme you can use them just as with any other utility in your HTML:
 
 ```html
-<h1 class="**text-2xs**">
-  Hello world!
-</h1>
-<span class="**text-10xl**">
+<span class="**text-2xs**">
   Hi world
 </span>
+<h1 class="**text-10xl**">
+  Hello world!
+</h1>
 ```
 
 In some cases, properties map to [variants](https://tailwindcss.com/docs/hover-focus-and-other-states) that can be placed in front of a utility to alter their behavior. For example, to add a `3xl` screen size that works just like the existing responsive screens, add a property under the `screens` key:
--- a/src/pages/docs/theme.mdx
+++ b/src/pages/docs/theme.mdx
@@ -207,30 +207,26 @@ Out of the box, your project will automatically inherit the values from [the def
 
 Most properties under `theme` map to utility classes and any values you add will become available as suffixes for existing utilities. If you'd like to preserve the default values for a theme option but also add new values, add your extensions under the `theme.extend` key in your configuration file. Values under this key are merged with existing `theme` values and automatically become available as new classes that you can use.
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/theme.mdx
```diff
@@ -205,24 +205,58 @@ Out of the box, your project will automatically inherit the values from [the def
...
+As an example, here we extend the `fontSize` property to add the classes `text-2xs` and `text-10xl` to change the font size:
+
+```js {{ filename: 'tailwind.config.js' }}
+/** @type {import('tailwindcss').Config} */
+module.exports = {
+  theme: {
+    extend: {
+      fontSize: {
+        '2xs': '0.625rem', // Adds a new `text-2xs` class
+        '10xl': '10rem', // Adds a new `text-10xl` class
```
reinink, line 219: I think it would be good to include line heights here, since we're extended the default font size scale, which [includes them](https://github.com/tailwindlabs/tailwindcss/blob/master/stubs/config.full.js#L324-L338).

```suggestion
        '2xs': ['0.625rem', '0.9rem'], // Adds a new `text-2xs` class
        '10xl': ['10rem', '1'], // Adds a new `text-10xl` class
```

src/pages/docs/theme.mdx
```diff
@@ -205,24 +205,58 @@ Out of the box, your project will automatically inherit the values from [the def
...
+
+After adding these to your theme you can use them just as with any other utility in your HTML:
+
+```html
+<span class="**text-2xs**">
+  Hi world
+</span>
+<h1 class="**text-10xl**">
+  Hello world!
+</h1>
```
reinink, line 234: Might be better to use the example text here as a way of explaining what's happening. Also call me crazy but I hate seeing an inline element at the same level a block, so maybe using a paragraph is better?
```suggestion
<h1 class="**text-10xl**">
  This is huge text
</h1>
<p class="**text-2xs**">
  This is tiny text
</p>
```

src/pages/docs/theme.mdx
```diff
@@ -205,24 +205,58 @@ Out of the box, your project will automatically inherit the values from [the def
...
+```
+
+In some cases, properties map to [variants](https://tailwindcss.com/docs/hover-focus-and-other-states) that can be placed in front of a utility to alter their behavior. For example, to add a `3xl` screen size that works just like the existing responsive screens, add a property under the `screens` key:
```
reinink, line 237: I know we haven't been 100% consistent with this, but we generally don't include the domain in our URLs, as it makes local development more annoying. You end up in a situation where you're making changes locally and pulling your hair out because nothing is updating in the browser, and then you realizing you're looking at the production site because you click a link that has the production domain in it 🤣
```suggestion
In some cases, properties map to [variants](/docs/hover-focus-and-other-states) that can be placed in front of a utility to alter their behavior. For example, to add a `3xl` screen size that works just like the existing responsive screens, add a property under the `screens` key:
```

PR #1671, opened 2023-09-06
docs: use 'create' command alias for creation of sveltekit project

the-evillevi:
[SvelteKit's documentation](https://kit.svelte.dev/docs/creating-a-project) uses `npm create svelte@latest my-app` instead of `npm init svelte@latest my-app`

`npm create` is an [alias](https://docs.npmjs.com/cli/v9/commands/npm-init?v=true) of `npm init`, so this change might not be necessary, but it might be beneficial to be coherent with SvelteKit's documentation.

Regards.

### DIFF
--- a/src/pages/docs/guides/sveltekit.js
+++ b/src/pages/docs/guides/sveltekit.js
@@ -18,7 +18,7 @@ let steps = [
     code: {
       name: 'Terminal',
       lang: 'terminal',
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Yep I think this is a good change. We already use `npm create` in some cases, e.g.: https://tailwindcss.com/docs/guides/vite

Thanks!

PR #1673, opened 2023-09-07
Move example below tailwind utilities

jonasdev:
If I remember correctly, you had to place custom CSS above utilities. This has changed, and in every other example on this page, the custom css is placed below. Should be the same for this example, to prevent confusion

### DIFF
--- a/src/pages/docs/adding-custom-styles.mdx
+++ b/src/pages/docs/adding-custom-styles.mdx
@@ -435,13 +435,12 @@ If you want to add some custom CSS that should always be included, add it to you
 ```css {{ filename: 'main.css' }}
 @tailwind base;
 @tailwind components;
+@tailwind utilities;
 
 /* This will always be included in your compiled CSS */
 .card {
   /* ... */
 }
-
-@tailwind utilities;
 ```
 
 Make sure to put your custom styles where they need to go to get the precedence behavior you want. In the example above, we've added the `.card` class before `@tailwind utilities` to make sure utilities can still override it.


jonasdev:
Okay, so right below, the description explains that;
````
Make sure to put your custom styles where they need to go to get the precedence behaviour you want. In the example above, we’ve added the .card class before @tailwind utilities to make sure utilities can still override it.
````

But then why start out by placing `.my-custom-style` below utilities in the very first example on the same page? Exactly what is confusing me.

https://tailwindcss.com/docs/adding-custom-styles#using-css-and-layer

reinink:
Hey! So I don't think this change makes sense, given that the description right below it explains why this custom CSS has been placed in that location.

The goal of this section is to explain how to remove unused custom CSS from your compiled CSS. You can use the `@layer` directive to achieve this for custom components, and when you're using the `@layer components` directive Tailwind will automatically put your CSS in the correct location.

However if you don't use the `@layer components` directive, then this custom CSS is always be included, but at that point you'll need to make sure it's placed in the correct order since it's not controlled by Tailwind.

> But then why start out by placing .my-custom-style below utilities in the very first example on the same page? Exactly what is confusing me. When to do what and what is preferred by default?

The point of that particular example is simply to explain that, when using Tailwind, it's totally fine to just include custom CSS in your stylesheet. We start with that simple example, and then go on to explain the `@layer` directive, and how that can be used to purge unused CSS and also determine the order of where the CSS is placed. We're basically just being careful not to explain too many new concepts at a time.

I hope that helps clear things up a little!

PR #1674, opened 2023-09-08
Add documentation for new variants

lukewarlow:
This adds documentation for the variants for the following media queries [`prefers-reduced-transparency`](https://github.com/tailwindlabs/tailwindcss/pull/11692), [`forced-colors`](https://github.com/tailwindlabs/tailwindcss/pull/11694), [`inverted-colors`](https://github.com/tailwindlabs/tailwindcss/pull/11693), and [`scripting`](https://github.com/tailwindlabs/tailwindcss/pull/11929).

### DIFF
--- a/src/pages/docs/hover-focus-and-other-states.mdx
+++ b/src/pages/docs/hover-focus-and-other-states.mdx
@@ -986,6 +986,30 @@ Tailwind also includes a `motion-safe` modifier that only adds styles when the u
 </button>
 ```
 
+### Prefers reduced transparency
+
+The `prefers-reduced-transparency` media query tells you if the user has requested that you minimize transparency.
+
+Use the `transparency-reduce` modifier to conditionally add styles when the user has requested reduced transparency:
+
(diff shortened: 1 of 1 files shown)


lukewarlow:
One thing to note I reckon it's worth linking out to the mdn pages for the various media queries.

src/pages/docs/hover-focus-and-other-states.mdx
```diff
@@ -986,6 +986,30 @@ Tailwind also includes a `motion-safe` modifier that only adds styles when the u
 </button>
 ```
 
+### Prefers reduced transparency
```
lukewarlow, line 989: Due to the fact this only works in Chrome i've not mentioned the `transparency-safe` modifier as that won't work across browsers.

I also think it's worth adding a note to say this is "experimental"?

sdavids: I suggest you add the matching documentation to the PR adding the feature instead of having a documentation PR for all four.

PR #1676, opened 2023-09-10
docs(background-position): fixed arbitrary values featured class

sorinraduta:


RobinMalfait:
Hey!

Going to close this because the `position` type hint shouldn't be necessary since `position` is the default for this arbitrary value: https://play.tailwindcss.com/DYdsvLvXoB

Thanks!

PR #1679, opened 2023-09-14
Link `ClassTable` properties to MDN docs

robclancy:
When I notice some new property I haven't used before but a coworker did I often find myself on tailwind docs and then googling mdn. I thought it would be useful to just be linked there.

This is just a link on the property with a dotted underline. I haven't really tested it and I don't know if it is something you would care for (personally I'd like `mdn`, `caniuse` links to the right of the CSS. Maybe even `caniuse` icons if they have an API, but I kept this simple just to show what I mean).

### DIFF
--- a/src/components/ClassTable.js
+++ b/src/components/ClassTable.js
@@ -26,8 +26,15 @@ function renderProperties(
           return (
             <Fragment key={i}>
               {'  '.repeat(indent)}
-              {property}: {transformedValue};
-              {px && <span className="text-indigo-400"> {`/* ${px} */`}</span>}
+              <a
+                className="border-b border-dotted"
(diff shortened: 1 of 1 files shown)


reinink:
Hey! So I think this is a neat idea, but if we're going to do this it's going to require a little more development, as there are places that this implementation doesn't currently work, such as the [box shadow colors](https://tailwindcss-com-git-fork-robclancy-master-tailwindlabs.vercel.app/docs/box-shadow-color) which link to `https://developer.mozilla.org/en-US/docs/Web/CSS/--tw-shadow-color`.

Given that we can't really prioritize this right now I am going to close this PR, but it's something we'll consider for the future – maybe the next iteration of the website.

Really appreciate you sharing the idea either way! 🙏

PR #1685, opened 2023-09-26
Update flex-grow.mdx

init-daemon:
"devDependencies": {
    "tailwindcss": "^3.3.3"
  }
with this version, grow is not working and...

I'm learning tailwindcss, I don't know if it's a bug or that my version of tailwind doesn't match the documentation I read(https://tailwindcss.com/docs/flex-grow #basic-usage) but I just know that mine doesn't work and I think my version is the right one. Grow doesn't work but flex-grow works like flex-grow css. I made my contribution to modifying the grows in the flex-grow documentation. I hope this improves something. please check the documentation

### DIFF
--- a/src/pages/docs/flex-grow.mdx
+++ b/src/pages/docs/flex-grow.mdx
@@ -9,8 +9,8 @@ import { HoverFocusAndOtherStates } from '@/components/HoverFocusAndOtherStates'
 
 export const classes = {
   utilities: {
-    '.grow': { 'flex-grow': '1' },
-    '.grow-0': { 'flex-grow': '0' },
+    '.flex-grow': { 'flex-grow': '1' },
+    '.flex-grow-0': { 'flex-grow': '0' },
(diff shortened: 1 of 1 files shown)


adamwathan:
Hey! `grow` is the correct class, it used to be `flex-grow` but we changed it quite a long time ago but kept `flex-grow` silently for backwards compatibility.

Can see here the expected CSS is generated:

image

https://play.tailwindcss.com/wOJA2Ip0GK

I'd recommend opening [a discussion](https://github.com/tailwindlabs/tailwindcss/discussions) that includes a reproduction project and someone can help you figure out why it's not working in your project 👍  My guess is just that you are using an old version of the framework.

init-daemon:
You're correct, thanks for your help ^.^ I've forgotten to launch "npx tailwindcss -i ./src/input.css -o ./dist/output.css --watch"
a stupid error,

PR #1689, opened 2023-09-27
Laravel Guide: Make including of vite resource more clear

rico-vz:
While the step description talks about both including the CSS in your head and styling using Tailwind, only the latter is highlighted. This PR solves this to provide more clarity to the reader by making the vite line also highlighted.


reinink:
Hey thanks! I think this is a reasonable change. I've tweaked the indentation and also applied the same change to the Laravel Mix guide for consistency 👍

PR #1692, opened 2023-10-02
Change variant used on `scroll-behavior` docs

Grienauer:
I think this help text is misleading for this component. at least the hover state makes no sense in my opinion and lead to confusions. maybe remove that for this component?

the html container will have always a hover state if someone clicks anywhere in the content of the page. so the first style will never apply? maybe I am incorrect. No hard feeling if the MR get's closed :)
At least I was quite confused and I think this is only added, because lots of the help-pages have the states. 
Thx!

### DIFF
--- a/src/pages/docs/overscroll-behavior.mdx
+++ b/src/pages/docs/overscroll-behavior.mdx
@@ -65,7 +65,7 @@ Use `overscroll-auto` to make it possible for the user to continue scrolling a p
 
 ### <Heading ignore>Hover, focus, and other states</Heading>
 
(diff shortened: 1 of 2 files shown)


thecrypticace:
Hey, so this example is at the bottom of every docs page so we can both explain how to apply utilities conditionally and point people to the documentation on how that feature works in general. We definitely want to keep the block to keep things consistent but you're right — changing `scroll-behavior` on hover doesn't make much sense. While it's still a bit of a contrived example I've updated this to use the `focus` variant instead.

Gonna wait for the build to complete so I can verify before I merge this.

PR #1693, opened 2023-10-03
Enable Next.js scroll restoration

### LINKED ISSUE #1686: Pages do not retain their scroll position.
Whenever I navigate to a previous page, the scroll position is not retained. It always starts at the top of the page.

thecrypticace:
Fixes #1686


thecrypticace:
This doesn't appear to 100% work when you go back multiple levels — or when fragments are involved _but_ this is far better than it is currently. Merging.

PR #1707, opened 2023-10-29
Add missing `g` flag in custom extraction example

ehaskins:
The example extraction regex is missing the `g` flag, so it currently only finds the fix match. Add `g` so it find all matches in `content`.


thecrypticace:
Thanks!

PR #1710, opened 2023-11-03
8.X renamed app.blade.php to welcome.blade.php

kzipp:


reinink:
Hey! So we intentionally used `app.blade.php` to represent a generic page template in your project as the `welcome.blade.php` file that ships with Laravel by default isn't intended to be a starting point for your own project, but rather it just provides a nice splash screen for when you first install Laravel.

Also worth mentioning that the `welcome.blade.php` file doesn't actually include your own compiled CSS (there is no `@vite('resources/css/app.css')` in this file), but actually includes a (limited) compiled version of Tailwind CSS [directly in the file](https://github.com/laravel/laravel/blob/40a7605dc1c5bea2d78fb31472e14b982252a5f2/resources/views/welcome.blade.php#L14-L16).

Because of that I think it's better to leave this as it is 👍

PR #1714, opened 2023-11-08
Fix custom opacity notation in opacity.mdx

DylanMerigaud:
image
image

######################################

image
image


thecrypticace:
Hey — this is not about arbitrary values so the current notation is correct and consistent with the rest of the docs.

PR #1715, opened 2023-11-18
Update aspect-ratio.mdx

zacharymichaelmoore:
curly brackets are incorrect

### DIFF
--- a/src/pages/docs/aspect-ratio.mdx
+++ b/src/pages/docs/aspect-ratio.mdx
@@ -14,7 +14,7 @@ export const classes = { utilities }
 
 ### Setting the aspect ratio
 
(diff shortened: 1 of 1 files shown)


adamwathan:
Hey! We use curly braces everywhere in all of the docs in this situation as a sort of placeholder for an actual value, for example:

image

So the idea on the aspect ratio page is that `aspect-{ratio}` represents things like `aspect-video` or `aspect-square`, not just the arbitrary values that use square brackets. Not perfect admittedly for sure, but going to keep it as-is for now for consistency, even if there's room for a better approach globally. Thanks anyways!

zacharymichaelmoore:
Do what you think is best. In this specific example it is incorrect. To set
the aspect ratio you do not use curly braces.

PR #1724, opened 2023-12-01
feat: add rspack guide for setup

lalalazero:
add rspack framework guide for tailwindcss installation


reinink:
Hey! Thanks for this contribution.

So I don't really know anything about Rspack or Rsbuild, but given that they look closely connected would it make sense to include these in a single guide somehow? Or if Rspack is the more common tool, maybe we just include that guide for now?

lalalazero:
Hi~ 

Rsbuild is at some level more like vite, intergrates many of features out-of-box while Rspack is more like the rust version of webpack, it needs many manual configurations. 

Yes, Rspack is more common tool, I'd like to keep rspack and remove rsbuild. Thanks for your advice.

src/pages/docs/guides/rspack.js
```diff
@@ -0,0 +1,321 @@
...
+        body: () => (
+          <p>
+            Install <code>tailwindcss</code> and its peer dependencies, then generate your{' '}
+            <code>tailwind.config.js</code> and <code>postcss.config.js</code> files.
+          </p>
+        ),
+        code: {
+          name: 'Terminal',
+          lang: 'terminal',
+          code: 'npm install -D tailwindcss postcss autoprefixer postcss-loader\nnpx tailwindcss init -p',
```
reinink, line 38: Is generating a `postcss.config.js` file still necessary with Rspack, given that you configure the PostCSS plugins in the `rspack.config.js` file?

lalalazero:
Hi~ thank you for testing it. In latest Rspack version, the `postcss.config.js` file is not needed.

reinink:
Hey @lalalazero, so testing this a little more this morning I learned that if you do use a `postcss.config.js` file you can actually omit the extra configuration from the `rspack.config.js` file.

This is nice because the `postcss.config.js` file is auto-generated for you using the `-p` flag. I've updated this guide to reflect this simpler setup.

Thanks for contributing this guide!

lalalazero:
Thank you!!! I appreciate your work very much!!

PR #1741, opened 2023-12-14
Update sveltekit install guide.

saadeghi:
In SvelteKit 2, `vitePreprocess` is no longer exported from `@sveltejs/kit/vite`

It should be imported from `@sveltejs/vite-plugin-svelte`

Source: https://kit.svelte.dev/docs/migrating-to-sveltekit-2#vitepreprocess-is-no-longer-exported-from-sveltejs-kit-vite


thecrypticace:
Thanks!

PR #1747, opened 2023-12-19
Fix blog heading hyperlink hover on dark mode

CalvinMoylan:
Currently, on a blog post when using dark mode, the [hyperlink button](https://tailwindcss.com/blog/tailwindcss-v3-4#dynamic-viewport-units) on hover text colour is the same as the background colour leading to the icon disappearing on hover.

This change just adds `dark:hover:text-slate-100` to the icon.

See below:
Current hover | neutral state | fixed hover
tailwind_hover_issue


reinink:
Thanks! Made some tweaks and merged it in 💪

PR #1761, opened 2024-01-09
update module export in Laravel guide

grrnikos:
Working with version 3.4.1 of Tailwind CSS, there is no `module.exports` in the `tailwind.config.js` file, but `export default` instead.


thecrypticace:
Thanks!

PR #1762, opened 2024-01-11
Fix invalid path of the output.css file

godnondsilva:
The changes done in PR #1754 has an issue where the output.css file is not detected by the HTML file.

This is because the href points to `/output.css`.

A better approach would be to add a relative path since all the three files (index.html, input.css and output.css) are within a single folder `src`.

I have added a (.) so that the relative path is taken into consideration.

Please review.


thecrypticace:
Thanks!

PR #1763, opened 2024-01-11
Document aria-busy variant

rik:
This was added back in Tailwind 3.3.3 (https://github.com/tailwindlabs/tailwindcss/pull/10966) but not documented.


reinink:
Nice, thanks @rik!

PR #1770, opened 2024-01-18
Update Angular logo on Framework Guides page

Hugo-Delattre:
The Angular logos on the Framework Guides are currently the old ones.

I've updated them to the new ones, while keeping the white logo version for the dark theme, and the colored logo version for the light theme.

- Before/After (white Angular logo, dark mode)

image
image

- Before/After (colored Angular logo, light mode)

image
image


thecrypticace:
Thanks! I've updated the light mode logo to the black one because:
- The angular brand guidelines express a preference for the black/white ones where color isn't absolutely necessary; AND
- Their color logo is not available as an SVG — I could probably create one from their illustrator file but I think the black and white logo is fine to use here.

PR #1773, opened 2024-01-24
Fix: typo on window variable

arung-agamani:
Renamed `Window.matchMedia()` to `window.matchMedia()` to give clarity that method of global variable `window` is being referred instead of `Window` interface instance method, and also analogous with the given example following the passage.


reinink:
Hey thanks for this suggestion. So technically I think what we have here is fine given that it's a pretty common convention to refer to interfaces and classes in this way (the `Window` interface in this case), even though the global instance variable is `window`. That said, I don't think this change hurts the legibility either, so happy to merge it in 👍

PR #1774, opened 2024-02-02
Clarify description of "truncate" and "text-ellipsis"

louisgthier:
In the Text Overflow documentation, "truncate" and "text-ellipsis" utilities have the same descriptions. This revision clarifies their differences.

### DIFF
--- a/src/pages/docs/text-overflow.mdx
+++ b/src/pages/docs/text-overflow.mdx
@@ -22,7 +22,7 @@ export const classes = {
 
 ### Truncate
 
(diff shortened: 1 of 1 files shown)


reinink:
Hey, thanks so much for this suggestion! You're right, it's kind of weird that these two utilities have the exact same description but do different things.

I've just made some tweaks to your PR — hope that's cool. I think the existing `text-ellipsis` description was actually fine, but I've updated the `truncate` description to explain that it prevents text from wrapping and also truncates overflowing text with an ellipsis if needed.

Thanks so much for pointing this out and for taking a crack at fixing this! 🙏

PR #1786, opened 2024-03-04
Fix showcase video playback which periodically fails to restart

wujingacross:
Fix the issue of frequent mouse movements in and out causing video playback to fail automatically

After playing the video once, it is easier to reproduce it by frequently moving it in and out again.

The state value has been fluctuating between leaving and looping.


reinink:
Hey thanks! I was able to reproduce this issue on Safari, and this seems to fix it, so thanks for this contribution!

PR #1798, opened 2024-03-19
Update Header.js to link to the correct header

onmax:


RobinMalfait:
Hey!

I already fixed it but missed your PR. I merged it anyway even though it's a no-op, but this way you still have some credit. Thanks for fixing!

PR #1799, opened 2024-03-20
💬 Update Remix Guide. Different config with Vite

ahaywood:
## Description

Recently, Remix migrated to Vite. With the update, the instructions within the Remix/Tailwind install guide no longer worked. 

Anytime I tried building the project, I'd get the following error:

CleanShot 2024-03-20 at 17 31 10

## Solution

I updated the docs to match the solution filed in this issue on Remix's side: https://github.com/remix-run/remix/issues/8832#issuecomment-1977696930

Key differences:
- Needs a postcss.config.js file
- Vite requires `?url` to respect the `.css` extension: `import stylesheet from "~/tailwind.css?url`

### DIFF
--- a/src/pages/docs/guides/remix.js
+++ b/src/pages/docs/guides/remix.js
@@ -77,7 +77,8 @@ let steps = [
     code: {
       name: 'root.tsx',
       lang: 'tsx',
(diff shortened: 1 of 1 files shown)


thecrypticace:
Thanks!

PR #1811, opened 2024-04-01
Resolve date discrepancy and remove need for tinytime dependency

brandonmcconnell:
This PR will…

1. **resolve a date mismatch issue caused by `tinytime` not preserving timezone information**
   
   I noticed this when sharing the job app links on Twitter. I'm not sure if this occurs in every time zone, or if I'm only seeing it because I'm currently traveling in Los Angeles.
   
   

2. **refactor and improve the `formatDate` function**
   
   1. **refactor and improve the `formatDate` function to build upon built-in APIs**
      
      Most of what tinytime was being used for can be offloaded to [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), which is natively supported with [great browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#browser_compatibility).
      
      The only regression/difference I noticed was that in the case of the job applications usage of `formatDate`, you lose the ordinal day number (e.g. `4` now instead of `4th`). If that's an important difference and you really want the ordinal number there, I have a helper that essentially shims `'ordinal` support for the day number. It would just be a deviation from the built-in `Intl.DateTimeFormatOptions` type, so if you do any TS type checking, you'll want to account for that as well.
      
   2. **removes the—now unused—the `tinytime` dependency**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,7 +55,6 @@
         "simple-functional-loader": "^1.2.1",
         "stringify-object": "^3.3.0",
         "tailwindcss": "^3.4.3",
-        "tinytime": "^0.2.6",
         "unified": "^10.1.2",
         "unist-util-filter": "^4.0.1",
         "unist-util-visit": "^4.1.2",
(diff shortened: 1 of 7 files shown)


brandonmcconnell:
Going to close this PR and give it some more thought, seeing as the Intl.DateTImeFormat might need to be imported or at least configured specifically for this Next.js env.

PR #1812, opened 2024-04-01
Resolve date discrepancy and remove tinytime dependency

brandonmcconnell:
👋🏼 **NOTE:** I am reopening this as a follow-up to #1811. I closed that one thinking that I might need to install another dependency to polyfill `Intl.DateTimeFormat` for server use, but that does not appear to be the case.

I'm unable to inspect the failures that caused the previous PR's build to fail, as I am not part of the `tailwindlabs` org on Vercel and do not have access to the `tailwindlabs/tailwindcss-com` resource.

If you report the failures here, I can take a stab at resolving them.

---

This PR will…

1. **resolve a date mismatch issue caused by `tinytime` not preserving timezone information**
   
   I noticed this when sharing the job app links on Twitter. I'm not sure if this occurs in every time zone, or if I'm only seeing it because I'm currently traveling in Los Angeles.
   
   

2. **refactor and improve the `formatDate` function**
   
   1. **refactor and improve the `formatDate` function to build upon built-in APIs**
      
      Most of what tinytime was being used for can be offloaded to [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), which is natively supported with [great browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#browser_compatibility).
      
      The only regression/difference I noticed was that in the case of the job applications usage of `formatDate`, you lose the ordinal day number (e.g. `4` now instead of `4th`). If that's an important difference and you really want the ordinal number there, I have a helper that essentially shims `'ordinal` support for the day number. It would just be a deviation from the built-in `Intl.DateTimeFormatOptions` type, so if you do any TS type checking, you'll want to account for that as well.
      
   2. **removes the—now unused—the `tinytime` dependency**

### DIFF
--- a/src/layouts/BlogPostLayout.js
+++ b/src/layouts/BlogPostLayout.js
@@ -6,6 +6,15 @@ import { MDXProvider } from '@mdx-js/react'
 import clsx from 'clsx'
 import Link from 'next/link'
 
+const dateFormat = {
+  weekday: 'long',
+  month: 'long',
+  day: 'numeric',
+  year: 'numeric',
+  hour: 'numeric',
+  minute: 'numeric',
+}
+
 export function BlogPostLayout({ children, meta }) {
   return (
     <div className="overflow-hidden">
@@ -50,7 +59,7 @@ export function BlogPostLayout({ children, meta }) {
                     className={clsx('absolute top-0 inset-x-0 text-slate-700 dark:text-slate-400')}
                   >
                     <time dateTime={meta.date}>
-                      {formatDate(meta.date, '{dddd}, {MMMM} {DD}, {YYYY}')}
+                      {formatDate(meta.date, dateFormat)}
                     </time>
                   </dd>
                 </dl>
--- a/src/layouts/BlogPostLayout.js
+++ b/src/layouts/BlogPostLayout.js
@@ -11,8 +11,6 @@ const dateFormat = {
   month: 'long',
   day: 'numeric',
   year: 'numeric',
-  hour: 'numeric',
-  minute: 'numeric',
 }
 
 export function BlogPostLayout({ children, meta }) {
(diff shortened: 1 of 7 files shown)


brandonmcconnell:
To clarify—as I didn't mention this before—I ran the build process locally, and everything built without any errors, and I can see the solution correctly rendering when running my fork locally:

adamwathan:
Thanks man! Here's a slice of the build errors, looks like this error is happening for most if not all blog posts:

image

thecrypticace:
@adamwathan  The `shortGeneric` error is because our Vercel build is still using Node 16 which we should change to be at least Node 18.

adamwathan:
Switched to Node 20 and triggered a redeploy, cross your fingers! 😄

src/layouts/BlogPostLayout.js
```diff
@@ -6,6 +6,15 @@ import { MDXProvider } from '@mdx-js/react'
 import clsx from 'clsx'
 import Link from 'next/link'
 
+const dateFormat = {
+  weekday: 'long',
+  month: 'long',
+  day: 'numeric',
+  year: 'numeric',
+  hour: 'numeric',
+  minute: 'numeric',
```
adamwathan, line 15: I think need to drop these in order to match the behavior of the live site:

image

brandonmcconnell: Ah yup! I updated those last and must've copied the wrong object. Updating that now.

brandonmcconnell: Should be resolved now via [42d3fde](https://github.com/tailwindlabs/tailwindcss.com/pull/1812/commits/42d3fde606cdc54065502aeaf85e296b7a6ce18d)

adamwathan:
Awesome thanks @brandonmcconnell!

PR #1814, opened 2024-04-10
Added overflow-clip examples

DavydeVries:
I noticed there were no `overflow-clip` examples, so I added them.

Screenshot/example on localhost:
tailwind-docs-overflow-clip


philipp-spiess:
@DavydeVries Hey! Really sorry this PR slipped through the cracks but we've in the meantime updated the website to a new v4 codebase and we aren't going to make changes to the old docs with new examples anymore :/ 

In general I am also not sure wether these examples are different enough from the existing `overflow-visible` and `overflow-hidden` examples that would justify their existence but I don't feel strongly about this.

PR #1819, opened 2024-04-20
Mention `motion-reduce` modifier under scroll behaviour

CTNicholas:
People will be looking for this one every time 👍

image


philipp-spiess:
@CTNicholas Sorry for the very late reply here but since we've moved to a new website for v4 we aren't adding new sections to the old v3 docs anymore. 

The scroll-behavior page currently has no example of how to use modifiers so this would still be a great addition if you add something like this `Applying on hover` section here: https://tailwindcss.com/docs/text-decoration-line#applying-on-hover. You'd have to create a new PR against the new `main` website branch though.

Issue #1816, opened 2024-04-13
[doc] Extend documentation on align-content


maurer2:
Hello, `align-content` for block and table elements recently gained support in all major browsers (see [caniuse for block elements]( https://caniuse.com/mdn-css_properties_align-content_block_context)), so it can potentially also be used to vertical align child elements in said elements. I prepared an example here: https://play.tailwindcss.com/RSyYxxieln

At the moment the documentation on `align-content`)[[content](https://tailwindcss.com/docs/align-content)] only mentions grid and flexbox, but it might make sense to mention, that most modern browsers now also support this property for the aforementioned elements as well.

Cheers

maurer2:
Okay, that makes sense. Thanks.

PR #1820, opened 2024-04-21
Extend intro section of the align-content page with a paragraph on support in block and table elements

maurer2:
Hello, following this [post from a couple of days ago](https://github.com/tailwindlabs/tailwindcss.com/issues/1816#issue-2241675480), I added another sentence to the intro section of the page on [align content](https://tailwindcss.com/docs/align-content) to highlight that this utility class is now also working on block- and table elements in most modern browsers:

More information:
* https://web.dev/blog/align-content-block
* https://developer.chrome.com/blog/align-content

Before
localhost_3000_docs_align-content (1)

After
localhost_3000_docs_align-content

Cheers


philipp-spiess:
Repeating my comment from #1816: Hey! Thanks for the contribution but I don't think we should mention that in the docs yet as we're still [supporting Safari 16.4](https://tailwindcss.com/docs/compatibility#browser-support) that doesn't have support for this yet. We'll consider it in a future itaration though 🙇

PR #1823, opened 2024-04-30
Update to latest Headless UI insiders release

thecrypticace:
Just some prep for Headless UI v2 stuff

### DIFF
--- a/src/components/Steps.js
+++ b/src/components/Steps.js
@@ -92,8 +92,6 @@ function CopyButton({ code }) {
         </svg>
       </button>
       <Transition
-        as="div"
-        className="absolute bottom-full left-1/2 mb-3.5 pb-1 -translate-x-1/2"
         show={state === 'copied'}
         enter="transform ease-out duration-200 transition origin-bottom"
         enterFrom="scale-95 translate-y-0.5 opacity-0"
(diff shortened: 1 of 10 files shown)


src/components/blog/ComboboxDemo.js
```diff
@@ -55,14 +55,18 @@ export default function Demo() {
             leaveTo="opacity-0"
             afterLeave={() => setQuery('')}
           >
-            <Combobox.Options className="absolute z-10 mt-1 max-h-60 w-full overflow-auto rounded-md bg-white py-1 text-base shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none sm:text-sm">
+            <Combobox.Options
+              as="ul"
```
RobinMalfait, line 59: Are these required? I know this was what we used to do, but I think we can just get rid of the ul/li entirely and use the default `<div>` elements.

I know we do this in Tailwind UI because we render additional `<li>` elements and they started to look funny without a parent `<ul>`.

src/components/Steps.js
```diff
@@ -92,6 +92,7 @@ function CopyButton({ code }) {
         </svg>
       </button>
       <Transition
+        as="div"
         className="absolute bottom-full left-1/2 mb-3.5 pb-1 -translate-x-1/2"
```
RobinMalfait: This is good, but another change we could do is move this directly to the div below. That way the `<Transition>` component only really exists for configuring the transitions themselves.

RobinMalfait: Oh hehe, you just did 💪

src/components/ThemeToggle.js
```diff
@@ -188,6 +188,7 @@ export function ThemeToggle({ panelClassName = 'mt-4' }) {
         </span>
       </Listbox.Button>
       <Listbox.Options
+        as="ul"
```
RobinMalfait, line 191: In this case it is required because of the `<li>` (unless we make that a normal `div` as well)

src/components/blog/ComboboxDemo.js
thecrypticace: I did this strictly because it's a demo from an existing blogpost. But yeah can update it — just have to verify styles and such but should be fine.

src/components/ThemeToggle.js
thecrypticace: Correct, I can _also_ make that change if you think we should

thecrypticace: Updated

RobinMalfait (approved):

PR #1827, opened 2024-05-20
Add more details to docs on why CSS variables are necessary for dynamic values

hkan:
Almost every day in Tailwind's official Discord server, someone asks why their classes are not working and their code snippet shows something like `h-[{someVariable}px]`. Explaining why this won't work in the documentation will be easier for telling people what they did wrong, by simply linking to this section of the document.

I'm not 100% confident on the wording, happy to apply suggestions.


hkan:
Noticed [this](https://tailwindcss.com/docs/content-configuration#dynamic-class-names) exists, which makes this change a duplicate. Closing.

PR #1837, opened 2024-06-26
Update ThemeReference.js to accurate text color css property

psullivan6:
- added a line to the description Array to ensure the `textColor` theme property is translated to `color` in the docs table, since `text-color` is not a valid CSS property


thecrypticace (approved):

thecrypticace:
Thanks!

PR #1844, opened 2024-07-17
Update table-layout.mdx

gcockroft:
changed 'class' to 'className'


adamwathan:
Hey! All of these are actually correct as `class` — even though the site is a React site the examples are vanilla HTML and our custom components for rendering this stuff set the HTML content directly so it needs to be `class` for things to work, not `className`. You can see here the demos are broken with this change:

image

So going to close this one.

PR #1864, opened 2024-08-19
fix: size-auto typo

andreasvirkus:
Minor typo fix


philipp-spiess:
Hey @andreasvirkus! This example works as expected since it shows how to _"remove an element's assigned width and height under a specific condition, like at a particular breakpoint"_ and the element already contains a `md:size-auto`. Thanks for the PR though 🙂

PR #1882, opened 2024-10-07
Using css variables update

crswll:
I think this approach is a lot clearer and provides some signigicant developer experience improvements:

* use any color format
* easier to reason about interopolation
* use a color picker

### DIFF
--- a/src/pages/docs/customizing-colors.mdx
+++ b/src/pages/docs/customizing-colors.mdx
@@ -350,9 +350,9 @@ First, define your colors any way you'd like.
 
   @layer base {
     :root {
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey Bill! Thanks for the change but we've since launched v4 of our website so we'll probably not going to update the legacy docs anymore. One of the features of v4 is that it is built from the ground-up with CSS variables and the `color-mix(…)` function so I think this might no longer be necessary but let us know if you think we can improve our new docs! Thanks a ton for the contribution and sorry for only getting to your PR now (we've been really busy with the v4 stuff these past months).

PR #1890, opened 2024-10-15
Update image path

muhammadsumon:


reinink:
Ahh good catch! I think it probably makes more sense to just include this image directly in this project — I've just updated this PR to do that. Thanks for this! 🤙

PR #1892, opened 2024-10-22
Note that `@layer` CSS at-rule will not be output.

kerryj89:
Returning to Tailwind after some time of using the native `@layer` system, I lost an hour of work thinking that something with my PostCSS setup was stripping my `@layer` at-rules.  I think the at-rule has matured to the point where people that haven't been using Tailwind for a while (or never had) might expect to see it output.  I see v4 will output this, but v3 simply reorders the rules for maximum compatibility.

I have read the documentation here multiple times and I feel if this statement was here it would click in my head sooner that  Tailwind v3 does not actually output the `@layer` at-rules and instead tracks and reorders them.  It was only when I started testing the outputs in the playground did I realize this distinction.

Feel free to reword and thanks for the otherwise great docs and css framework!

### DIFF
--- a/src/pages/docs/functions-and-directives.mdx
+++ b/src/pages/docs/functions-and-directives.mdx
@@ -83,6 +83,8 @@ Any custom CSS added to a layer will only be included in the final build if that
 
 Wrapping any custom CSS with `@layer` also makes it possible to use modifiers with those rules, like `hover:` and `focus:` or responsive modifiers like `md:` and `lg:`.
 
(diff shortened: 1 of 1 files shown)


kerryj89:
If anyone stumbles on this, here's a helpful article that talks about what goes on under the hood on how Tailwind handles @layer https://emamoah.medium.com/using-native-css-cascade-layers-with-tailwindcss-cb11af266104

PR #1899, opened 2024-11-07
Added Inertia guide with AdonisJS

HowWhat42:
As asked in the AdonisJS community, I added the steps for Inertia guide


philipp-spiess:
Hey! Thanks for the contribution but this seems Interia is optional with AdonisJS so I think it's better to omit this from the default guides to not convolute the example. If that's something that lots of people will look for I think we could maybe add a separate installation guide for this setup but I doesn't seem highly requested I'm afraid. So going to close this for now, thanks though!

PR #1905, opened 2024-11-15
feat: add documentation section for vite setup

WellingtonNico:
Hello guys, I just created a new react vite project to play with tailwind, I followed the documentation but it was not enough to make things working, so I added what was missing in the docs.

great job, be happy!

### DIFF
--- a/src/pages/docs/guides/vite.js
+++ b/src/pages/docs/guides/vite.js
@@ -63,6 +63,26 @@ let tabs = [
         },
       },
       {
+        title: 'Add css section to vite config',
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey! Configuring a PostCSS setup separately to the `postcss.config.js` file that `npx tailwindcss@3 init -p` should not have been necessary. Did you run the init command back then? 

Nevertheless we are no longer recommending the use of Vite with PostCSS since the release of Tailwind CSS v4 and the new `@tailwindcss/vite` plugin: https://tailwindcss.com/docs/upgrade-guide#using-vite

PR #1908, opened 2024-11-21
[DOC] Installation | use `export default` in tailwind.config.js

egyjs:
match the current `npx init` command


philipp-spiess (approved):
Thank you!

PR #1910, opened 2024-11-22
Fix typo in v4 stacked variants migration example

jitinl:
Fix a typo in the v4 migration docs where the v3's `last:*:pb-0` should be `*:last:pb-0` but is (incorrectly) written as `*:first:pb-0`


philipp-spiess (approved):
Thanks!

PR #1917, opened 2024-11-28
remove z-index on beam

danhollick:
Before:
Screenshot 2024-11-28 at 14 09 23

After:
Screenshot 2024-11-28 at 14 09 06

No idea why this class was there tbh


philipp-spiess (approved):
Looks good!

PR #1918, opened 2024-11-30
Update docs for Svelte 5

arithefirst:
Recently, Svelte version 5 released introducing the runes system. Part of this included changing the `<slot/>` tag to `{@render children()}`. This PR makes the [SvelteKit install guide](https://tailwindcss.com/docs/guides/sveltekit) reflect those changes.

### Before
```svelte
<script>
  import "../app.css";
</script>

<slot />
```

### Proposed Change
```svelte
<script>
  import "../app.css";
  let { children } = $props();
</script>

{@render children()}
```

### DIFF
--- a/src/pages/docs/guides/sveltekit.js
+++ b/src/pages/docs/guides/sveltekit.js
@@ -107,9 +107,10 @@ let steps = [
       lang: 'html',
       code: `<script>
   import "../app.css";
+  let { children } = $props();
 </script>
 
-<slot />`,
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey! Thanks for the contribution and sorry for not getting to it faster. We've in the meantime launched a new Tailwind CSS website which now includes this change too. Because of this, we're going to leave the old site as-is for now.

PR #1921, opened 2024-12-06
Typo fix: Change `theme()` to `var()`

nuotsu:
Minor typo in the docs:

Should be `var(--breakpoint-xl)` not `theme(--breakpoint-xl)`

Screenshot 2024-12-05 at 7 00 53 PM


philipp-spiess:
@nuotsu Unfortunately `var()` does not work within `@media` queries. That's exactly the use case of why we still expose the `theme()` function since it gets replaced to the inlined theme value at compile time.

PR #1933, opened 2025-01-04
Add a feature to copy class names to the clipboard

IZUMIRU:
I’ve added a feature that allows copying class names. It's basically the same code as the Customizing Colors page.

- https://github.com/tailwindlabs/tailwindcss.com/blob/2f727dc474f06a82649fbba7c98fd3eab9389980/src/components/ColorPaletteReference.js

I personally find this a very helpful addition, and I would greatly appreciate it if you would consider merging it.
Thank you for your time and consideration!

- ref: #1660

https://github.com/user-attachments/assets/9740a060-96e9-41da-ba15-c8165ef5ec4a

### DIFF
--- a/src/components/ClassTable.js
+++ b/src/components/ClassTable.js
@@ -3,6 +3,7 @@ import { isObject } from '@/utils/isObject'
 import { castArray } from '@/utils/castArray'
 import clsx from 'clsx'
 import { Heading } from '@/components/Heading'
+import { Transition } from '@headlessui/react'
 
 function renderProperties(
   properties,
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey! With the change to the v4 website and these class names now being more generic (i.e you'll have to replace `<number>` with your desired number, I don't think that's something that makes sense anymore. Sorry for the unfortunate timing here as we've rewritten the website right while you were working on your contribution.

IZUMIRU:
Understood. I appreciate your great contribution!

IZUMIRU:
@philipp-spiess 
I’d like to have the copy feature in v4 as well. Would it be okay if I contribute?

philipp-spiess:
@IZUMIRU I do think it'd be confusing with class names like `ml-<number>` which aren't going to be useful if copied into your clipboard so currently thinking that we probably won't accept a copy feature here I'm afraid

PR #1936, opened 2025-01-08
Removing unused components

davidturissini:
There are a few components that do not appear to be referenced anywhere in the code. This PR removes the unused components. Running `npm run build` in the repo still results in a successful build.

These components were discovered via `$ npx remove-unused@latest`.

Check https://removeunused.com/ for some more info


philipp-spiess:
This is awesome thanks for helping us but we went a little further and did a complete rewrite of the website which deletes even more stuff 🙈 Because of the move to the new website, I don't think we'll want to make any big non-necessary updates to the old docs anymore, but still really appreciate it!

PR #1937, opened 2025-01-17
Fix font size utilities reference namespace

valtism:
Font size is controlled by the `--text-*` namespace, not `--font-*` which controls font-family


RobinMalfait:
Hey!

Thanks for the PR, but now that Tailwind CSS v4 is released, we redirect the v4 beta docs to the "normal" docs so this change won't make a difference right now. At the time of writing these beta docs, the namespace was still `--font`, but now that should be corrected to `--text` in the normal docs as well.

https://tailwindcss.com/docs/theme#theme-variable-namespaces

I appreciate the PR!

PR #1939, opened 2025-01-19
Improve tailwindcss.com docs for Vite config

visrut7:
Problem:
- Tailwind doesn't work out of the box with Vite after I followed the guide from tailwindcss.com website, but I was able to work around it after I came across this answer: https://stackoverflow.com/questions/77741027/how-to-use-tailwindcss-with-vite-not-react-or-something-just-vanilla-js

- So I am adding that step in the docs

### DIFF
--- a/src/pages/docs/guides/vite.js
+++ b/src/pages/docs/guides/vite.js
@@ -104,6 +104,31 @@ let tabs = [
   }`,
         },
       },
+      {
+        title: 'Configure Vite',
+        body: () => (
+          <p>
+            Update your <code>vite.config.js</code> file to include Tailwind CSS in the PostCSS
+            plugins.
+          </p>
+        ),
+        code: {
+          name: 'vite.config.js',
+          lang: 'js',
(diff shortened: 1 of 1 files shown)


henrikbjorn:
Not needed see `https://vite.dev/guide/features#postcss`

It already loads `postcss` so it seems you didn't configure postcss correctly.

visrut7:
I followed this guide step by step: https://tailwindcss.com/docs/guides/vite

where we also install `postcss` in step 2, after that tailwind was not working so had to look it up and then found a solution on StackOverflow.

henrikbjorn:
Works for me,

Sure you didn't install postcss after running `init` for tailwind ? In that case I think it wouldn't create the `postcss.config.js` file.

philipp-spiess:
Hey! Thanks for the contribution. We've recently launched our new v4 website which contains a brand new Vite plugin and new reworked Vite guides that should be clearer 👍

PR #1941, opened 2025-01-21
Fix type error in docs

jimhigson:
content extraction example was not valid according to tailwind's typscript types. The type is:

```ts
type ExtractorFn = (content: string) => string[]
```

which means that this is not allowed:

```ts
extract: {
      wtf: (content) => {
        return content.match(/[^<>"'`\s]*/g)
      }
    }
```

Ie, `ExtractorFn` does not allow null, but `Regexp.prototype.match` is typed as retuning `string[] | null`.


philipp-spiess:
Going to close this one as we've migrated to a new docs site 👍  Thanks though

PR #1945, opened 2025-01-23
feat: ability to copy colors in v4 docs

ryanhaticus:
- Enables the ability to copy colors in the v4 `/docs/colors` page.
- Converts `oklch` to hex as it's more recognizable from a DX perspective. This should be okay because the color palette doesn't utilize values.

Gif:
click-to-copy

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "@tailwindcss/postcss": "4.0.0",
     "@types/mdx": "^2.0.13",
     "clsx": "^2.1.1",
-    "colorizr": "^3.0.7",
     "dedent": "^1.5.3",
     "fathom-client": "^3.7.2",
     "feed": "^4.2.2",
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "@tailwindcss/postcss": "4.0.0",
     "@types/mdx": "^2.0.13",
     "clsx": "^2.1.1",
+    "colorizr": "^3.0.7",
     "dedent": "^1.5.3",
     "fathom-client": "^3.7.2",
     "feed": "^4.2.2",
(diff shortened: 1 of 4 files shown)


tristonarmstrong (commented):
Dude said, "let there be tabs"

DerpyDinosaur:
Nice looks great 👍🏻

tristonarmstrong:
much better ;*

ryanhaticus:
@philipp-spiess, thoughts on this quality of life change? Perhaps someone with more knowledge regarding Headless UI can fix the issues @tristonarmstrong mentioned.

victormihalache:
> As a side node: The conversion to hex values here is probably not a good idea. I understand that hex codes are more common but the reason we used `oklch(…)` is because with Tailwind CSS v4 we are using on a wide-gamut color palette that often contains colors [_outside of the rgb spectrum_](https://tailwindcss.com/blog/tailwindcss-v4#modernized-p3-color-palette). Because the color space of `oklch` is bigger than that of `rgb`, it means that converting from one into the other is _loosing information_ and as a result of that you will not end up with the same color.

It would be were it not for the fact that when using Figma or Sketch you basically _have_ to use RGB. As far as I know the reason the click-to-copy functionality is useful is because you'd try out different colors whilst designing, rather than actually using it in code (you'd just use Tailwind directly in that case).


Screenshot 2025-01-28 at 12 47 13 AM
Screenshot 2025-01-28 at 12 47 34 AM

victormihalache:
Awesome! Also, doesn't copy on Safari (does in Firefox).


https://github.com/user-attachments/assets/bf0e9bf6-8448-47a5-ab76-36399b69ae69

philipp-spiess:
Posted an update in the issue thread: https://github.com/tailwindlabs/tailwindcss.com/issues/1956#issuecomment-2674122202

PR #1947, opened 2025-01-23
Fix typo in theme page example

heappynd:
Updated the CSS variable from `var(--bg-mint-500)` to `var(--color-mint-500)` in the div element.


reinink:
Thanks!

PR #1948, opened 2025-01-23
Changes `@import` paths to be consumer-facing

wongjn:
Especially fixes the code example for disabling Preflight.


reinink:
Thanks Justin!

PR #1953, opened 2025-01-23
fix: updated the margin logical properties example code

Rees1993:
Page: [margin - Spacing](https://tailwindcss.com/docs/margin#using-logical-properties)

Fixes the margin logical properties example where it doesn't reflect the difference when `dir="rtl"` is set. The example also utilises the `ms-8` & `me-8` classes.


reinink:
Good catch, thanks! 🤙

PR #1958, opened 2025-01-23
fix(docs): add missing bg class for dark variant on hover

DeepakBalaraman:
This pr adds the missing class name `dark:hover:file:bg-violet-700` in the [::file](https://tailwindcss.com/docs/hover-focus-and-other-states#file) example provided in the documentation.


reinink:
Good catch, thanks! 🤙

PR #1961, opened 2025-01-23
Prefer `@custom-variant` over `@variant`

wongjn:
(Sorry if I misinterpreted the intentions of `@custom-variant` and `@variant`)

- Replaces code usages of `@variant` to `@custom-variant` as per tailwindlabs/tailwindcss#15663.
- Disambiguates some usage of `@variant` in examples to show that it should be used inside a rule, and not at the root.


thecrypticace:
💯 

Thanks!

PR #1968, opened 2025-01-23
Increase size for search button to improve accessibility on mobile

alexanderbluhm:
Adds the same size for the search button on the header as the menu button next to it.

I searched for other occurrences of `SearchButton` but it's used with it's own padding in all other places, so I didn't change the `SearchButton` component directly.

## Before:
Screenshot 2025-01-23 at 22 54 18@2x
Screenshot 2025-01-23 at 22 56 11@2x

## After:
Screenshot 2025-01-23 at 22 55 32@2x
Screenshot 2025-01-23 at 22 57 47@2x


reinink:
Good catch, thanks! 🤙

PR #1976, opened 2025-01-24
Clarification on relative paths

PhilETaylor:
Time wasted? 1hour. 

Hope this clarification helps someone else avoid the rabbit hole I went down

I'm also an idiot, so happy if this gets rejected as such also.


philipp-spiess (approved):
Thanks! Reworded this a bit though so we don't need a new paragraph 👍

PR #1986, opened 2025-01-24
Update sveltekit.tsx - to svelte 5 syntax

dei8bit:
In Svelte 5, content can be passed to components in the form of snippets and rendered using render tags.

In legacy mode, content inside component tags is considered slotted content, which can be rendered by the component using a <slot> element

https://svelte.dev/docs/svelte/legacy-slots


philipp-spiess (approved):
Oh nice great catch, you're absolutely right!

PR #1990, opened 2025-01-24
show `bunx` alternative

browner12:
show popular `bunx` alternative to `npx`

wondering if there'd be some interest in showing some love to `bun` and `bunx` alternatives, as they have become a quite popular alternative.

just doing this one spot right now as an example, but would make make adjustments if there's interest.


philipp-spiess:
Hey! Thanks for the idea but there are a lot of package managers these days and we’d rather leave that to the reader to decide while only providing a generic default so we are likely not going to change this.

PR #1991, opened 2025-01-24
show laravel vite plugin import

browner12:
to help indicate that both the Laravel AND Tailwind Vite plugins should be installed, and that the Tailwind plugin does not replace the Laravel one.


philipp-spiess:
Hey! Appreciate the PR but it's unclear to me if this is really clearer, especially since the import is unused in the code snippet. The main thing this codeblock should show that you add a new import and use it in the plugins array. I think we'll going to leave this as-is for now

PR #1992, opened 2025-01-24
Replace usage of `theme()` to `var()` as recommended by Tailwind v4

Blaumaus:
At the [Using the theme() function](https://tailwindcss.com/docs/upgrade-guide#using-the-theme-function) section docs advice to use `var()` over `theme()`, but there is a part of code where `theme()` is still used


philipp-spiess (approved):

philipp-spiess:
Oh, thanks a ton!

PR #1995, opened 2025-01-24
Update Header.js in `v2` to mention `v4` (instead of `v3`) 

flx-sta:
Fix link to latest verson in v2 documentation


philipp-spiess:
This doesn't seem to build on Vercel anymore. :( I think we'd first have to upgrade all dependencies of v2 to work with a more recent Node.js version which doesn't seem worth it to do I'm afraid.

PR #2007, opened 2025-01-25
Blog page tweaks

brabli:
Fix a couple of bits on an older blog post and add a newline on v4 blog post to better separate CSS blocks


src/blog/tailwindcss-v3-2/index.mdx
philipp-spiess, line 1: Can you revert the changes here? If I check out your branch and run prettier on this file, it will actually print the initial version.

philipp-spiess (changes requested):
Hey! Thanks for the contribution but the changes in the 3.2 blog post are actually unexpected when when I run prettier on this file in your branch it actually reverts them. 

I can't push on this branch of yours to fix it so have to ask you to please revert that change in `src/blog/tailwindcss-v3-2/index.mdx`.

brabli:
I ran prettier myself and it looks like it's only affecting the link elements I inlined, presumably they're a bit long for prettier's liking. I've removed that change but I re-added the missing quotes as they didn't appear to be affected by the autoformatting, does it work on your end now?

philipp-spiess (approved):

philipp-spiess:
Thanks!

PR #2008, opened 2025-01-26
Update @tailwind/cli installation guide to include config loading

le-michael:
Updated the installation docs since you need to explicitly define the config location now. I believe the old version just loaded the config automatically.

Example of loading from tests:
https://github.com/tailwindlabs/tailwindcss/blob/7e20c3b58750f769d9b6e35bf283689c86341b1d/integrations/cli/config.test.ts#L31


philipp-spiess:
Hey! Thanks for the contribution but a JavaScript config file is no longer required and it's recommended you migrate to your config to CSS: https://tailwindcss.com/docs/upgrade-guide

PR #2012, opened 2025-01-27
Remove duplicate svgs

danhollick:
Cleaned up the icons on the landing page so that they are responsive to theme changes, rather than swapping out the whole svg.


RobinMalfait (approved):
😍

PR #2013, opened 2025-01-27
Fix page jump to bottom on safari

### LINKED ISSUE #2004: Bug in quick search
### How to reproduce?

- In Safari click on quick search (or press cmd + k) and without typing anything click outside the modal
- The page gets scrolled to the bottom

https://github.com/user-attachments/assets/afd6f788-6086-4765-af2c-3ffbd8b672c6

### Environment

Browser: Safari (Version - 18.2)

spudoodle:
#### Reference Issues / PRs

closes #2004

#### What does this implement/fix?

Fixes the quick jump to bottom of the page in safari

This is an existing bug in docsearch: https://github.com/algolia/docsearch/issues/1260

#### Any other comments?

No

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -37,7 +37,7 @@ export default function Home() {
           <FooterMeta className="px-4 md:px-6 lg:px-8" />
         </div>
       </div>
-      <div className="!fixed" aria-hidden="true">
+      <div className="fixed!" aria-hidden="true">
         <input type="text" tabIndex={-1} />
       </div>
     </div>
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -37,6 +37,9 @@ export default function Home() {
           <FooterMeta className="px-4 md:px-6 lg:px-8" />
         </div>
       </div>
+      <div className="!fixed" aria-hidden="true">
+        <input type="text" tabIndex={-1} />
+      </div>
     </div>
   );
 }


src/app/page.tsx
```diff
@@ -37,6 +37,9 @@ export default function Home() {
           <FooterMeta className="px-4 md:px-6 lg:px-8" />
         </div>
       </div>
+      <div className="!fixed" aria-hidden="true">
```
RobinMalfait, line 40:
```suggestion
      <div className="fixed!" aria-hidden="true">
```

RobinMalfait:
Beautiful, thank you very much!

I updated the class to be `fixed!` instead of `!fixed` because this is the v4 way of applying `!important`. Thanks!

RobinMalfait (approved):

spudoodle:
Hey @RobinMalfait, I just realized that the 'https://tailwindcss.com/docs/' URL utilizes group routes. So our current fix does not target this route, and the previously reported issue persists on this page

philipp-spiess:
@spudoodle Should we put this fix into `src/app/layout.tsx` instead?

spudoodle:
@philipp-spiess Yes, I tried out transferring the fix to `src/app/layout.tsx` before the body tag, and it works as intended. I also added 'disabled' to the input element.

https://github.com/user-attachments/assets/2263df0c-781a-4417-bf30-e6151694ca91

PR #2021, opened 2025-01-28
fix: correct syntax for flex custom properties

redoy49:
Replaced flex-[<custom-property>] with flex-(<custom-property>) to align with the intended syntax and ensure consistency across the codebase.
fix


redoy49:
fix: correct syntax for flex custom properties

philipp-spiess:
Good catch! thanks

Issue #2028, opened 2025-01-28
`!important` docs gone from v4


stephenlrandall:
In v3, there's: https://v3.tailwindcss.com/docs/configuration#important-modifier.

In v4, no such section exists anymore. As far as I can tell, there's no documentation anywhere in the v4 docs for the "!" prefix.

georg-stone:
@stephenlrandall It seems there are docs now.

PR #2030, opened 2025-01-29
Document the `important` option

RobinMalfait:
This PR adds documentation for the `important` option on the `@import "tailwindcss";` at-rule.

Remaining questions:

- [ ] Is this the correct spot to document this? (We redirect `/docs/configuration` to `/docs/theme`, so not 100% sure)
- [ ] Should we document a solution for the `#app` selector strategy that is mentioned in the v3 docs?

Fixes: https://github.com/tailwindlabs/tailwindcss.com/issues/2028

### DIFF
--- a/src/docs/theme.mdx
+++ b/src/docs/theme.mdx
@@ -531,39 +531,6 @@ Then you can use `@import` to include your theme variables in other projects:
 
 You can put shared theme variables like this in their own package in monorepo setups or even publish them to NPM and import them just like any other third-party CSS files.
 
-### Important
-
-The `important` option can be used to make all utilities have a 
-
-To generate utilities as `!important`, set the important option on the Tailwind
-CSS import:
-
-```css
-/* [!code filename:app.css] */
-/* [!code word:important] */
-@import "tailwindcss" important;
-```
-
-To generate individual utilities as `!important`, use the `!` syntax at the end of a utility class:
-
-<CodeExampleStack>
-
-```html
-<!-- [!code filename:index.html] -->
-<!-- [!code word:!] -->
-<div class="bg-red-500!"></div>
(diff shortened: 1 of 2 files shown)


src/docs/theme.mdx
```diff
@@ -530,6 +531,39 @@ Then you can use `@import` to include your theme variables in other projects:
 
 You can put shared theme variables like this in their own package in monorepo setups or even publish them to NPM and import them just like any other third-party CSS files.
 
+### Important
+
+The `important` option can be used to make all utilities have a
```
philipp-spiess, line 536:
```suggestion
```

src/docs/theme.mdx
```diff
@@ -530,6 +531,39 @@ Then you can use `@import` to include your theme variables in other projects:
...
+
+The `important` option can be used to make all utilities have a 
+
+To generate utilities as `!important`, set the important option on the Tailwind
+CSS import:
+
+```css
+/* [!code filename:app.css] */
+/* [!code word:important] */
+@import "tailwindcss" important;
```
philipp-spiess, line 544: I wonder if we should also explain where to put that `important` if you de-sugar that import (i think it's on the `@import "tailwindcss/utilities.css"` right?)

src/docs/theme.mdx
```diff
@@ -530,6 +531,39 @@ Then you can use `@import` to include your theme variables in other projects:
...
+```
+
+To generate individual utilities as `!important`, use the `!` syntax at the end of a utility class:
```
obecker, line 547: Since this has changed from v3 (where `!` always appeared at the beginning of a utility class), this should be mentioned on the `upgrade` page as well.

PR #2031, opened 2025-01-29
Fix `padding` documentation

AmProsius:
The `padding` docs say that e.g. `p-px` translates to `padding: calc(var(--spacing) * 1px);` which is not true.

As with the `margin` utilities, `p-px` translates to `padding: 1px;`.


philipp-spiess:
Thanks for the PR! I also fixed `space-*-px` while at it. I somehow messed up committing onto your PR though so this went straight into `main` haha sorry! It does contain a commit attributed to you so not going to try to fix this. Thanks again!

PR #2035, opened 2025-01-30
Add example for dark/light mode automatic color switching

XavierGeerinck:
This PR adds the ability to automatically switch colors based on Dark/Light mode. This is a commonly used "feature" as many people are using ShadCN nowadays. This also resolve the discussion: https://github.com/tailwindlabs/tailwindcss/discussions/13211


philipp-spiess:
Hey there! Thanks for the PR but we're not going to include this for now. While I agree that swapping variables at runtime can make dark mode a bit easier, I worry that this example is too convoluted and will rather be confusing to people. We might include something smaller in the future though.

PR #2038, opened 2025-01-30
Fix `adding-custom-styles` documentation

sangmin4208:
Correct CSS variable names in modifiers example

- Changed `--font-size-*` to `--text-*`
- Changed `--line-height-*` to `--leading-*`

https://tailwindcss.com/docs/adding-custom-styles#modifiers


RobinMalfait (approved):
Oh good catch, thanks!

PR #2041, opened 2025-01-31
Document upgrade path for `--config` CLI option

RobinMalfait:
This PR updates the upgrade guide for when you are using the Tailwind CLI with the `--config` option.

This option doesn't exist anymore, and instead you should use the `@config` directive in your CSS.

### DIFF
--- a/src/docs/upgrade-guide.mdx
+++ b/src/docs/upgrade-guide.mdx
@@ -1,3 +1,5 @@
+import { CodeExampleStack } from "@/components/code-example";
+
 export const title = "Upgrade guide";
 export const description = "Upgrading your Tailwind CSS projects from v3 to v4.";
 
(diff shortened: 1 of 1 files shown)


src/docs/upgrade-guide.mdx
```diff
@@ -74,6 +76,25 @@ npx tailwindcss -i input.css -o output.css
 npx @tailwindcss/cli -i input.css -o output.css
 ```
 
+#### CLI options
+
+In v3, you could use the `--config` option to point to your configuration file.
+In v4 there is no configuration file anymore. If you still require a
+configuration file, make sure to include it in your CSS file and use the `@config` directive:
```
thecrypticace, line 83: What do you think about something like this? (probably still needs to be split across multiple lines tho)

```suggestion
In v3, you could use the `--config` option to point to your configuration file. Now that projects are configured via CSS this option is no longer available. For projects that require a configuration file for compatibility reasons, load the file by using the `@config` directive in your CSS:
```

src/docs/upgrade-guide.mdx
```diff
@@ -74,6 +76,25 @@ npx tailwindcss -i input.css -o output.css
...
+<CodeExampleStack>
+
+```css
+/* [!code filename:input.css] */
+@config "./tailwind.config.js";
+```
+
+```shell
+npx @tailwindcss/cli -i input.css -o output.css
+```
```
thecrypticace, line 94: Should you include the config file content as an example code block too?

RobinMalfait:
Closed it for now, will document it properly if more people run into it.

PR #2042, opened 2025-01-31
Add footer to showcase and blog pages

danhollick:


src/app/blog/[slug]/page.tsx
```diff
@@ -116,6 +117,13 @@ export default async function DocPage(props: Props) {
               </div>
             </section>
           </GridContainer>
+          <GridContainer className="mt-46">
+            <FooterSitemap className="*:first:border-l-0 *:last:border-r-0" />
+          </GridContainer>
+
+          <div>
+            <FooterMeta className="px-4 md:px-6 lg:px-8" />
+          </div>
```
Comment (reinink) [inline src/app/blog/[slug]/page.tsx:126]: I haven't tested this, but is this wrapping `<div>` actually required? (it's in a couple other places in this PR too).

Comment (danhollick) [reply src/app/blog/[slug]/page.tsx]: Ah, no good catch. It's used as a layout wrapper elsewhere and I should have removed it when I removed the styles.

PR #2048, opened 2025-02-03
Fix incorrect custom key in columns.mdx

paulhennell:
The page incorrectly references the container key on the columns page.

### DIFF
--- a/src/docs/columns.mdx
+++ b/src/docs/columns.mdx
@@ -303,7 +303,7 @@ Learn more about the gap utilities in the [gap documentation](/docs/gap).
 
 <CustomizingYourTheme
   utility="columns"
-  themeKey="container"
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey! Thanks for the PR but this documentation is on purpose. We consider the `--columns-*`-specific namespace more of a backward-compatibility only option. 

For new customizations, it's better to use `--container-*` since that will be shared across different, conceptually similar utilities (e.g. [container queries](https://tailwindcss.com/docs/responsive-design#container-queries). You can see that the `columns` utility looks up both namespaces: https://github.com/tailwindlabs/tailwindcss/blob/ac202ffd50e6009982da4d31db6e2aea3600666d/packages/tailwindcss/src/utilities.ts#L1700

PR #2049, opened 2025-02-03
doc: add migration steps

steveoh:
it was not clear to me that the migration tool would not perform this migration. The docs make it sounds like the choice is to use the migration tool or perform these manual steps. But even after running the migration tool, these steps are still recommended.

### DIFF
--- a/src/docs/upgrade-guide.mdx
+++ b/src/docs/upgrade-guide.mdx
@@ -62,6 +62,8 @@ export default defineConfig({
 });
 ```
 
+Migrating from PostCSS is as simple as removing the postcss.config.js file and removing the autoprefixer and postcss packages.
+
 ### Using Tailwind CLI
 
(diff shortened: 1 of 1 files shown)


philipp-spiess:
@steveoh Hey! Thanks for the PR. The reason why the upgrade tool doesn't migrate to the Vite extension and the reason why we don't go into detail in the docs  is that there are many different ways people use Vite (e.g. standalone or via a meta-framework etc.) and also many different ways people have set up their PostCSS setups before. E.g. your update mentions deleting the postcss plugin which is only there if you run the upgrade tool before and deleting postcss/autoprefixer, both of which aren't necessary in Vite if you've used the postcss plugin before (postcss is added automatically by Vite for example).

I think keeping this section a bit more vague on purpose helps since the users is ultimately the person who knows best about their setup so going to leave this as-is for now but we'll keep thinking on how to make this easier. Thanks!

PR #2050, opened 2025-02-03
fix: close bracket ] in background-attachment.mdx

Helias:


RobinMalfait:
Oh good catch, thanks!

PR #2055, opened 2025-02-05
`background-size` arbitrary value type hint

wongjn:
The example would emit `background-position` instead of `background-size`. Although the type-hint would not be needed if you used `auto` as one of the two values, I've added the type-hint unilaterally to keep things simple.

| Before | After |
| --- | --- |
|  image |  image |
| image |  image | 

Instigated by https://github.com/tailwindlabs/tailwindcss/issues/15880#issuecomment-2635917784

### DIFF
--- a/src/components/content.tsx
+++ b/src/components/content.tsx
@@ -51,6 +51,7 @@ export function UsingACustomValue({
   value,
   variable,
   dataType,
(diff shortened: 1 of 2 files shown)


subproject22:
It should be `size` not `lenght`

PR #2056, opened 2025-02-05
Fix `create astro` link

HiDeoo:
The first step of the [Astro framework guide](https://tailwindcss.com/docs/installation/framework-guides/astro) links to the Astro documentation for `create astro` using a link to the docs landing page and an anchor that doesn't exist on this page.

After discussing with the Astro docs team, this PR updates the link to point to https://docs.astro.build/en/install-and-setup/#install-from-the-cli-wizard which is a more appropriate link for the `create astro` command.


RobinMalfait (approved):
Thank you!

PR #2068, opened 2025-02-06
fix: css link in Vite installation dist -> src

messenjer:
It should be `src` instead of `dist`

After building with `vite build`, the output in `dist/` might look like:

`<link href="/assets/styles-abc123.css" rel="stylesheet">`

Thanks for the new v4 release, it works like a charm! 🔥


philipp-spiess (approved):
Oh, thanks for catching this!

PR #2077, opened 2025-02-09
[chore]: use install as dev dependencies for cli

### LINKED ISSUE #2074: v4 Getting Started Tailwind CLI - suggest installing as dev dependencies instead of regular dependencies
I recreated the tiny example project used to demonstrate [setting up a project with the CLI tool](https://tailwindcss.com/docs/installation/tailwind-cli) with one difference: I ran a different command for step 01. Instead of using `npm install tailwindcss @tailwindcss/cli`, I used `npm install -D tailwindcss @tailwindcss/cli`. Since the Tailwind code isn't going to be included in a project that uses Tailwind (only the output CSS that Tailwind generates), it seems more appropriate as a dev dependency as opposed to a regular dependency.

SleeplessOne1917:
Closes #2074 

Very tiny change. As stated in the issue, these don't need to be production dependencies.

### DIFF
--- a/src/app/(docs)/docs/installation/(tabs)/tailwind-cli/page.tsx
+++ b/src/app/(docs)/docs/installation/(tabs)/tailwind-cli/page.tsx
@@ -27,7 +27,7 @@ const steps: Step[] = [
     code: {
       name: "Terminal",
       lang: "shell",
-      code: "npm install tailwindcss @tailwindcss/cli",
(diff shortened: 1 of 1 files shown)


philipp-spiess (commented):
Hey! Appreciate the comment but we deliberately removed the flag so that the reader can decide what makes more sense for their setup so going to close this for now.

PR #2078, opened 2025-02-10
fix: text in accent-color input's label

Jungho-Cheon:
In the opacity setting of the accent color, the black color that is not in the code in the example is being used.
I modified it to purple just like the example code.
- AS-IS
as-is
- TO-BE
to-be


philipp-spiess (approved):
Thank you!

PR #2083, opened 2025-02-10
Update adding-custom-styles.mdx

dukesteen:
Add a link to the MDN docs on CSS types


philipp-spiess (commented):
Appreciate the PR but I do think this is more confusing. While we are looking at the docs for the custom utilities syntax, we unfortunately do only support a few of the data types (those mentioned in the docs). So linking to a comprehensive list of CSS types might suggest we support more than that which we don't, yet!

PR #2088, opened 2025-02-13
Fix order of Vite plugins in SvelteKit installation guide

tetra-fox:
The order of Vite plugins matter, as they are run sequentially.
I encountered an issue where classes applied conditionally with the [class: directive](https://svelte.dev/docs/svelte/class#The-class:-directive) would not work unless the `tailwindcss` plugin was run after the `sveltekit` plugin.
This PR reorders the plugins array to correct this behavior.

### DIFF
--- a/src/app/(docs)/docs/installation/framework-guides/sveltekit.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/sveltekit.tsx
@@ -63,9 +63,9 @@ export let steps: Step[] = [
 
         export default defineConfig({
           plugins: [
+            sveltekit(),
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey! The order is on purpose, Tailwind CSS must come before the Svelte hook: https://github.com/tailwindlabs/tailwindcss/issues/14809#issuecomment-2469920472 
Do you have a repro for the issue you were seeing? The transform hook in Tailwind CSS _only ever changes CSS files_ and should have 0 effect on `.svelte` files.

PR #2089, opened 2025-02-14
Fix object-none example in docs

Curs3W4ll:
The doc about the `object-fit` property seems to display a wrong example of the `object-none` class.

The example image seems to be stretched instead of being displayed at its original size.

# Example

Screenshot 2025-02-15 at 00 03 39

# Original image
Screenshot 2025-02-15 at 00 04 00


philipp-spiess (approved):
Thanks!

PR #2094, opened 2025-02-16
fix(docs): install tailwind as a dev dependency

otabekshoyimov:
On v4 vite plugin docs doesn't include --save-dev command during install, is this a mistake or a breaking change from v3 to v4? 

Is tailwind a runtime dependency now?

If not, I changed the docs of vite plugin install page to include --save-dev command

Screenshot 2025-02-16 at 16 21 07

### DIFF
--- a/src/app/(docs)/docs/installation/(tabs)/using-vite/page.tsx
+++ b/src/app/(docs)/docs/installation/(tabs)/using-vite/page.tsx
@@ -28,7 +28,7 @@ const steps: Step[] = [
       name: "Terminal",
       lang: "shell",
       code: dedent`
-        npm install tailwindcss @tailwindcss/vite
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey! You're definitely right, Tailwind CSS is not a _runtime_ dependency in the sense that it has a runtime in your client builds but we deliberately removed the `--dev` flags from the docs so that the reader can decide what makes more sense for their setup since we found that there is no clear standard of runtime vs. dev dependencies since most apps are bundled anyways and not published as npm packages (the only place where there's a real distinction between the two). E.g. in a project I worked on we used non-dev dependencies as a way to indicate which libraries we need to attribute in the license agreement and since Tailwind does produce CSS that you ship in your app, we did had to include it. 

Going to close this for now since it's a deliberate decision but thank you so much 👍

PR #2098, opened 2025-02-19
Add missing type hints in usage examples

wongjn:
Some *Using a custom value* sections are missing type hints for CSS variable usage examples. This PR adds them in.


philipp-spiess (approved):
Thank you!!

PR #2104, opened 2025-02-20
fix: add `cursor-pointer` to buttons in header

shun-shobon:
I noticed that the version selector dropdown menu and the button to open the search modal in the header of the v4 documentation do not have cursor-pointer applied.
Since these elements are clickable, the cursor should change to provide a better user experience.
Therefore, I added cursor-pointer to ensure the cursor changes appropriately.

### DIFF
--- a/src/components/header.tsx
+++ b/src/components/header.tsx
@@ -53,7 +53,7 @@ function VersionPicker() {
   return (
     <Menu>
       <MenuButton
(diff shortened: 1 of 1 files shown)


philipp-spiess (commented):
Thanks but we changed this on purpose to be closer to how browsers render native `<button>` elements. The change comes backed into Tailwind CSS v4 too: https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor

PR #2107, opened 2025-02-21
Add a paragraph on breakpoint sorting

### LINKED ISSUE #2099: [v4] Wrong specificity of media query with a custom breakpoint value
**What version of Tailwind CSS are you using?**

v4.0.7

**What build tool (or framework if it abstracts the build tool) are you using?**

`@tailwindcss/cli` v4.0.7

**What version of Node.js are you using?**

v22.14.0

**What browser are you using?**

Chrome

**What operating system are you using?**

macOS

**Reproduction URL**

```
$ git clone git@github.com:mr-wildcard/tailwindcss-v4-bug-ordered-custom-breakpoint-value.git
$ cd tailwindcss-v4-bug-ordered-custom-breakpoint-value
$ npm i
$ npx @tailwindcss/cli -i ./input.css -o ./output.css
```

**Describe your issue**

If we customize the value of a breakpoint using `px` unit:
```css
@theme {
    --breakpoint-2xl: 1600px;
}
```

the generated media query for that specific breakpoint end up **before** the other media queries using `rem`:
Image

This is problematic in case we provide a value that is wider than the other breakpoints using `rem`: in that case we expect the `2xl` variant styles to be applied last but they've in fact less specificity than the other media queries:
Image

Image

---

I believe that the unit of the value is also sorted because if I use the `vh` unit, the generated media query is placed at the end, but if I use `ch` unit, it's placed before.

Not sure if that's really an issue and how it could be fixed, but I think that we can quickly fall into this kind of trap and lose hours debugging.
Maybe add a warning in the documentation?

philipp-spiess:
This has been causing repeated confusion already so let's add a section to clarify this.

Closes #2099

### DIFF
--- a/src/docs/responsive-design.mdx
+++ b/src/docs/responsive-design.mdx
@@ -324,6 +324,8 @@ This updates the `2xl` breakpoint to use `100rem` instead of the default `96rem`
 </div>
 ```
 
+Utilities using these breakpoints are automatically sorted based on the value of the breakpoint, as long as the units match.
+
 Learn more about customizing your theme in the [theme documentation](/docs/theme).
 
 ### Removing default breakpoints
--- a/src/docs/responsive-design.mdx
+++ b/src/docs/responsive-design.mdx
@@ -324,7 +324,9 @@ This updates the `2xl` breakpoint to use `100rem` instead of the default `96rem`
 </div>
 ```
 
-Utilities using these breakpoints are automatically sorted based on the value of the breakpoint, as long as the units match.
+Note that it's important to always use the same unit for defining your breakpoints or the generated utilities may be sorted in an unexpected order, causing breakpoint classes to override each other in unexpected ways.
+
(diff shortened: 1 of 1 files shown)


src/docs/responsive-design.mdx
```diff
@@ -324,6 +324,8 @@ This updates the `2xl` breakpoint to use `100rem` instead of the default `96rem`
 </div>
 ```
 
+Utilities using these breakpoints are automatically sorted based on the value of the breakpoint, as long as the units match.
```
adamwathan, line 327:
```suggestion
Note that it's important to always use the same unit for defining your breakpoints or the generated utilities may be sorted in an unexpected order, causing breakpoint classes to override each other in unexpected ways.

Tailwind uses `rem` for the default breakpoints, so if you are adding additional breakpoints to the defaults, make sure you use `rem` as well.
```

PR #2109, opened 2025-02-24
Fix missing color: on doc example

samtidball:
Fixed a missing `color:` on a `&:hover` example in: https://tailwindcss.com/docs/colors#referencing-in-css


philipp-spiess (approved):
Thank you!

Issue #2105, opened 2025-02-21
website CSS issue


lorypelli:
this happens when you reload for a second:
![Image](https://github.com/user-attachments/assets/044f4820-4c1f-4b50-8c37-a8d4b1bf0a78)

philipp-spiess:
Hey! Can't really reproduce this, does this happen all the time for you?

PR #2110, opened 2025-02-24
Fix color flash on page load when setting dark/light mode

Gazler:
Previously, when loading the landing page, the page colors would flash briefly. For light mode the background flashes black before the white background loads.

For dark mode, the header would flash white, and the text black, before the correct colors render.

This was happening because the script was being executed after the page had finished loading, likely due to how the page is packaged into chunks.

To fix this, the dark mode script is now set inline, so will be part of the generated HTML.

This fixes https://github.com/tailwindlabs/tailwindcss.com/issues/2105

Here's a video showing what was happening before.

https://github.com/user-attachments/assets/2019277f-d947-4289-bf41-89fef7aa3c26

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -159,8 +159,7 @@ export default async function RootLayout({
         <meta name="application-name" content="Tailwind CSS" />
         <meta name="msapplication-TileColor" content="#38bdf8" />
         <meta name="msapplication-config" content={v("/favicons/browserconfig.xml")} />
-
-        <Script src={`data:text/javascript;base64,${btoa(darkModeScript)}`} />
+        <script type="text/javascript" dangerouslySetInnerHTML={{__html: darkModeScript}}></script>
       </head>
       <body>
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Thank you!

Gazler:
FWIW, it looks like that issue you are seeing only happens locally? Perhaps it's Next.js taking over the error page?

https://tailwindcss-com-git-fork-gazler-fix-dark-mo-afe988-tailwindlabs.vercel.app/404 shows correctly for me.

Gazler:
I see. In that case, the issue still exists on the 404 page. https://tailwindcss.com/not-found 

Definitely much less of an issue than it happening on every other page though!

philipp-spiess:
You mean the flash? Yeah I don't know how else we can fix it here since synchronous scripts just don't evaluate. But I think this is reasonable too to have this issue on not-found pages. There aren't accessed too often thankfully 👍

PR #2115, opened 2025-02-26
add overflow auto to fix show more expand list

Hlool24:
add overflow-y-auto to fix show more expand list when it overflow the content over the footer

image

### DIFF
--- a/src/app/(docs)/layout.tsx
+++ b/src/app/(docs)/layout.tsx
@@ -18,7 +18,7 @@ export default async function Layout({
           <DocsSidebar />
         </MobileNav>
       </div>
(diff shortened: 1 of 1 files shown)


philipp-spiess:
@Hlool24 Hey! Can you explain what exactly this is supposed to fix? I am not able to get into the status that you showed in your screenshot because the API table pushing the footer further down 🤔

Hlool24:
@philipp-spiess it fixes the overflow of the page if there is a large list of utilities and you click show more it will be overflowed under the footer like in this page https://tailwindcss.com/docs/border-radius

philipp-spiess:
@Hlool24 But if you have a large list of utilities and click show-more, we only render more items and push the rest of the website down? I don't see any overflow? What am I missing? 😅  

https://github.com/user-attachments/assets/0a561611-19f1-4189-8c44-3dfa6b85d243

Hlool24:
it looks like it only shows in my device and only in chrome, i have tested it in different devices and it worked well

philipp-spiess:
@Hlool24 Yeah that's indeed very strange. Could it be either a caching issue or some browser-extensions interfering? 🤔 Nevertheless I don't think this is something that we should fix in the code for everyone though, so I'm going to close this PR for now. Thank you!

Issue #2134, opened 2025-03-10
Inaccurate folder structure referenced in installation setup


xkjjx:
From what I understand, installing Tailwind v4 in Ruby on Rails uses the `tailwind/application.css` file, not the `application.tailwind.css` file. In addition, the file comes prefilled with `@import "tailwindcss";`.

The [main installation guide](https://tailwindcss.com/docs/installation/framework-guides/ruby-on-rails) doesn't provide the correct information, however.

philipp-spiess:
This should also be fixed by your PR, ty!

PR #2135, opened 2025-03-10
Remoeved unnecessary/incorrect information about installation

xkjjx:
Addresses #2134


philipp-spiess (approved):
Ah nice, thank you! I ran the setup guide locally but you're right the `@import "tailwidncss"` part is now inserted for you 👍

PR #2138, opened 2025-03-10
Documents `overflow-wrap` utilties

danhollick:
Also removes `break-words` from the `word-break` docs, although it will still work for backwards compatibility.

### DIFF
--- a/src/docs/overflow-wrap.mdx
+++ b/src/docs/overflow-wrap.mdx
@@ -1,0 +1,99 @@
+import { ApiTable } from "@/components/api-table.tsx";
+import { Example } from "@/components/example.tsx";
+import { Figure } from "@/components/figure.tsx";
+import { ResponsiveDesign } from "@/components/content.tsx";
+
+export const title = "overflow-wrap";
+export const description = "Utilities for controlling line breaks in an overflowing element.";
+
+<ApiTable
+  rows={[
+    ["wrap-normal", "overflow-wrap: normal;"],
+    ["wrap-anywhere", "overflow-wrap: anywhere;"],
+    ["wrap-break-word", "overflow-wrap: break-word;"],
+  ]}
+/>
+
+## Examples
+
+### Normal
+
+Use the `wrap-normal` utility to only allow line breaks at hard wrap opportunities, like spaces and hyphens:
+
+<Figure>
+
+<Example padding={false}>
+  {
+    <p className="mx-auto max-w-xs border-x border-x-pink-400/30 py-8 [overflow-wrap:normal] text-gray-900 dark:text-gray-200">
+      The longest word in any of the major English language dictionaries is{" "}
+      <span className="font-bold">pneumonoultramicroscopicsilicovolcanoconiosis,</span> a word that refers to a lung
+      disease contracted from the inhalation of very fine silica particles, specifically from a volcano; medically, it
+      is the same as silicosis.
+    </p>
+  }
+</Example>
+
+```html
+<!-- [!code classes:wrap-normal] -->
+<p class="wrap-normal">The longest word in any of the major...</p>
+```
+
+</Figure>
+
+### Break Word
+
+Use the `wrap-break-word` utility to allow line breaks at soft wrap opportunities, like between letters in a word, if needed:
+
+<Figure>
+
+<Example padding={false}>
+  {
+    <p className="mx-auto w-min max-w-xs border-x border-x-pink-400/30 py-8 text-gray-900 dark:text-gray-200">
+      The longest word in any of the major English language dictionaries is{" "}
+      <span className="font-bold [overflow-wrap:break-word]">pneumonoultramicroscopicsilicovolcanoconiosis,</span> a
+      word that refers to a lung disease contracted from the inhalation of very fine silica particles, specifically from
+      a volcano; medically, it is the same as silicosis.
+    </p>
+  }
+</Example>
+
+```html
+<!-- [!code classes:wrap-break-word] -->
+<p class="w-min">
+  The longest word... <span class="wrap-break-word font-bold">pneumonoultramicroscopicsilicovolcanoconiosis,</span>
+</p>
+```
+
+</Figure>
+
+### Anywhere
+
+Use the `wrap-anywhere` utility to allow line breaks at soft wrap opportunities, like between letters in a word, if needed:
+
+<Figure>
+
+<Example padding={false} className="max-h-96">
+  {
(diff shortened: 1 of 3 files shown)


src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,99 @@
...
+import { ResponsiveDesign } from "@/components/content.tsx";
+
+export const title = "overflow-wrap";
+export const description = "Utilities for controlling line breaks in an overflowing element.";
+
+<ApiTable
+  rows={[
+    ["wrap-normal", "overflow-wrap: normal;"],
+    ["wrap-anywhere", "overflow-wrap: anywhere;"],
+    ["wrap-break-word", "overflow-wrap: break-word;"],
```
reinink, line 13: Suggested new order for these. I think it would be nice to have this match the order that they are explained on the page, and right now "break-word" comes before "anywhere". Plus I think we should move "normal" to be the last thing on the page, since it's really a reset (ie. on a breakpoint).

```suggestion
    ["wrap-break-word", "overflow-wrap: break-word;"],
    ["wrap-anywhere", "overflow-wrap: anywhere;"],
    ["wrap-normal", "overflow-wrap: normal;"],
```

src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,99 @@
...
+  ]}
+/>
+
+## Examples
+
+### Normal
```
reinink, line 19: Generally Adam and I have tried to come up with section titles that describe what you're trying to accomplish. Admittedly we haven't done that everywhere (it's really hard in some places), but I think for new pages it would be good to try and do that.


```suggestion
### Wrapping normally
```

src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,99 @@
...
+```html
+<!-- [!code classes:wrap-break-word] -->
+<p class="w-min">
+  The longest word... <span class="font-bold wrap-break-word">pneumonoultramicroscopicsilicovolcanoconiosis,</span>
+</p>
+```
+
+</Figure>
+
+### Anywhere
```
reinink, line 69:
```suggestion
### Wrapping anywhere
```

src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,99 @@
...
+      refers to a lung disease contracted from the inhalation of very fine silica particles, specifically from a
+      volcano; medically, it is the same as silicosis.
+    </p>
+  }
+</Example>
+
+```html
+<!-- [!code classes:wrap-break-word] -->
+<p class="w-min">
+  The longest word... <span class="font-bold wrap-break-word">pneumonoultramicroscopicsilicovolcanoconiosis,</span>
```
reinink, line 63: I think we can omit "font-bold" from the snippet, since we don't do that on other pages that have this word in bold.

And maybe we should add "..." to the end?

```suggestion
  The longest word... <span class="wrap-break-word">pneumonoultramicroscopicsilicovolcanoconiosis,</span> ...
```

src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,99 @@
...
+      to a lung disease contracted from the inhalation of very fine silica particles, specifically from a volcano;
+      medically, it is the same as silicosis.
+    </p>
+  }
+</Example>
+
+```html
+<!-- [!code classes:wrap-anywhere] -->
+<p class="w-min">
+  The longest word... <span class="font-bold wrap-anywhere">pneumonoultramicroscopicsilicovolcanoconiosis,</span>
```
reinink, line 89: As noted elsewhere, I think we can remove the "font-bold" here, and maybe add "..." to the end?

```suggestion
  The longest word... <span class="wrap-anywhere">pneumonoultramicroscopicsilicovolcanoconiosis,</span> ...
```

src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,99 @@
...
+</p>
+```
+
+</Figure>
+
+Unlike the `wrap-break-word`, `wrap-anywhere` allows the browser to factor in mid-word line breaks when calculating the intrinsic `min-content` size of the element.
+
+### Responsive design
+
+<ResponsiveDesign element="p" property="wrap" defaultClass="normal" featuredClass="break-word" />
```
reinink, line 99: The property needs to be `overflow-wrap` here, and the classes are missing the `wrap-` prefix.

```suggestion
<ResponsiveDesign element="p" property="overflow-wrap" defaultClass="wrap-normal" featuredClass="wrap-break-word" />
```

reinink (commented):
Back to you, Dan! 🤙

reinink (changes requested):
Oops, I didn't select the "Request changes" option.

src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,99 @@
+import { ApiTable } from "@/components/api-table.tsx";
+import { Example } from "@/components/example.tsx";
+import { Figure } from "@/components/figure.tsx";
+import { ResponsiveDesign } from "@/components/content.tsx";
+
+export const title = "overflow-wrap";
+export const description = "Utilities for controlling line breaks in an overflowing element.";
```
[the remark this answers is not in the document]
danhollick: yeah i think so.

src/docs/overflow-wrap.mdx
```diff
@@ -0,0 +1,97 @@
...
+import { ResponsiveDesign } from "@/components/content.tsx";
+
+export const title = "overflow-wrap";
+export const description = "Utilities for controlling line breaks within words in an overflowing element.";
+
+<ApiTable
+  rows={[
+    ["wrap-normal", "overflow-wrap: normal;"],
+    ["wrap-break-word", "overflow-wrap: break-word;"],
+    ["wrap-anywhere", "overflow-wrap: anywhere;"],
```
reinink, line 13: Let's change the order to:

1. `wrap-normal`
2. `wrap-break-word`
3. `wrap-anywhere`

And update the sections on the page to match this order.

src/docs/word-break.mdx
```diff
@@ -9,7 +9,6 @@ export const description = "Utilities for controlling word breaks in an element.
 <ApiTable
   rows={[
     ["break-normal", "overflow-wrap: normal;\nword-break: normal;"],
```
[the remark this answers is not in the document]
reinink: Would be good to also move `break-normal` and the "Normal" example to be last.

PR #2140, opened 2025-03-11
Document `pointer-*` and `any-pointer-*` variants

danhollick:
Here's what the example looks like when a fine pointer is used:
Screenshot 2025-03-11 at 11 21 11

Here's what the example looks like when a coarse pointer is used:
Screenshot 2025-03-11 at 11 21 27

### DIFF
--- a/src/docs/hover-focus-and-other-states.mdx
+++ b/src/docs/hover-focus-and-other-states.mdx
@@ -4113,7 +4113,7 @@ A quick reference table of every single variant included in Tailwind by default.
       </tr>
       <tr>
         <td>
-          <a href="#orientation" className="whitespace-nowrap">
+          <a href="#pointer-and-any-pointer" className="whitespace-nowrap">
             pointer-fine
           </a>
         </td>
@@ -4123,7 +4123,7 @@ A quick reference table of every single variant included in Tailwind by default.
       </tr>
       <tr>
         <td>
-          <a href="#orientation" className="whitespace-nowrap">
+          <a href="#pointer-and-any-pointer" className="whitespace-nowrap">
             pointer-coarse
           </a>
         </td>
@@ -4133,7 +4133,7 @@ A quick reference table of every single variant included in Tailwind by default.
       </tr>
       <tr>
         <td>
-          <a href="#orientation" className="whitespace-nowrap">
+          <a href="#pointer-and-any-pointer" className="whitespace-nowrap">
             pointer-none
           </a>
         </td>
@@ -4143,7 +4143,7 @@ A quick reference table of every single variant included in Tailwind by default.
       </tr>
       <tr>
         <td>
-          <a href="#orientation" className="whitespace-nowrap">
+          <a href="#pointer-and-any-pointer" className="whitespace-nowrap">
             any-pointer-fine
           </a>
         </td>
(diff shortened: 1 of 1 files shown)


src/docs/hover-focus-and-other-states.mdx
```diff
@@ -1857,6 +1857,76 @@ Use the `not-forced-colors` variant to apply styles based when the user is _not_
...
+    </fieldset>
+  }
+</Example>
+
+```html
+<!-- [!code classes:pointer-coarse:mt-6,pointer-coarse:grid-cols-3,pointer-coarse:gap-4,pointer-coarse:p-4] -->
+<fieldset aria-label="Choose a memory option">
+  <div class="flex items-center justify-between">
+    <div>RAM</div>
+    <a> See performance specs </a>
```
reinink, line 1912: Just some extra spacing here. I think we generally show the `href="#"` in our snippets too.

```suggestion
    <a href="#">See performance specs</a>
```

reinink (changes requested):
Couple minor thing 👀

src/docs/hover-focus-and-other-states.mdx
```diff
@@ -2341,14 +2411,14 @@ Alternatively, you can configure shortcuts for common data attributes you're usi
...
+@custom-variant has-checked (&[data-ui~="checked"]);
 ```
 
 You can then use these custom `data-*` variants in your project:
 
 ```html
-<!-- [!code classes:data-checked:underline] -->
-<div data-ui="checked active" class="data-checked:underline">
+<!-- [!code classes:has-checked:underline] -->
+<div data-ui="checked active" class="has-checked:underline">
```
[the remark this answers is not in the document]
danhollick: ummm, no. wtf? Must have `cmd`+ `d`ed a couple too many times

src/docs/hover-focus-and-other-states.mdx
```diff
@@ -1857,6 +1857,76 @@ Use the `not-forced-colors` variant to apply styles based when the user is _not_
 
 Tailwind also includes a [forced color adjust](/docs/forced-color-adjust) utilities to opt in and out of forced colors.
 
+### pointer and any-pointer
```
reinink, line 1860: We should update our current "Registering a custom variant" example, since it's now baked in. Maybe do something that has two states, like "hover" and "focus" could be a "hocus" variant (we've used this as an example in the past).

src/docs/hover-focus-and-other-states.mdx
```diff
@@ -2884,24 +2954,14 @@ With at-rule custom variants the `&` placeholder isn't necessary, just like when
 If you find yourself using the same arbitrary variant multiple times in your project, it might be worth creating a custom variant:
 
 ```css
-@custom-variant pointer-coarse {
-  @media (pointer: coarse) {
-    @slot;
-  }
-}
+@custom-variant theme-midnight (&:where([data-theme="midnight"] *));
```
danhollick, line 2957: @reinink not quite sure if this is right. I removed the bit about shorthand syntax because it doesn't seem relevant with this example

reinink: I just updated this whole section to match what I've done in #2182 👍

PR #2141, opened 2025-03-11
Document `scripting-*` variants

danhollick:


src/docs/hover-focus-and-other-states.mdx
```diff
@@ -1873,6 +1873,17 @@ Use the `portrait` and `landscape` variants to conditionally add styles when the
 </div>
 ```
 
+### scripting
+
+Use the `scripting`, `scripting-none`, and `scripting-initial` variants to conditionally add styles based whether the user has scriping, such as JavaScript, enabled:
+
+```html
+<!-- [!code classes:scripting-none:block] -->
+<div class="hidden scripting-none:block">
```
reinink, line 1882: I wonder, is this better?

```suggestion
<!-- [!code classes:scripting:hidden] -->
<div class="scripting:hidden">
```

reinink (changes requested):
One minor suggestion.

src/docs/hover-focus-and-other-states.mdx
```diff
@@ -4089,6 +4100,36 @@ A quick reference table of every single variant included in Tailwind by default.
...
+        </td>
+      </tr>
+      <tr>
+        <td>
+          <a href="#scripting" className="whitespace-nowrap">
+            scripting-initial
+          </a>
+        </td>
+        <td>
+          <code className="whitespace-nowrap before:content-none after:content-none">@media (scripting: initial)</code>
```
reinink, line 4120: The value here is `initial-only`:

```suggestion
          <code className="whitespace-nowrap before:content-none after:content-none">@media (scripting: initial-only)</code>
```

src/docs/hover-focus-and-other-states.mdx
```diff
@@ -1873,6 +1873,17 @@ Use the `portrait` and `landscape` variants to conditionally add styles when the
 </div>
 ```
 
+### scripting
+
+Use the `scripting`, `scripting-none`, and `scripting-initial` variants to conditionally add styles based whether the user has scriping, such as JavaScript, enabled:
```
reinink, line 1878: We're removing the `scripting` and `scripting-initial` variants, and we're replacing `scripting-none` with `noscript`. This is more inline with the `print` and `forced-colors`.

PR #2148, opened 2025-03-13
Update angular.tsx

USSDlover:
- Add step to install required types for editors intellisense to work
- Use `@use` keyword insteadl of `@import` due to this warning `Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.`


adamwathan:
Hey! Going to close this because the typography plugin is not required at all, that's an optional plugin you can install if you need it, and we also discourage using Sass with Tailwind:

https://tailwindcss.com/docs/compatibility#sass-less-and-stylus

…so using `@import` is correct for the recommended path.

PR #2151, opened 2025-03-14
Missing colon and preposition

chuntington:
This PR adds a colon where one would be expected after a code description, and also adds a preposition "to" achieve proper grammar.


philipp-spiess (approved):
Thank you!

PR #2152, opened 2025-03-14
Add link to v4 docs in content-configuration.mdx

alexsapps:
For folks very stuck on content configuration who might be using tailwind v4, add mention of the new docs and clarify that v3 config does not work. A lot of Q/A sites / online resources still only talk about v3 config without mentioning it is for v3 only and v4 docs aren't highly ranked, and attempting to use v3 config does not cause any error. I updated a few questions on stackoverflow as well after spending several hours trying various tooling, path patterns and options to build my packages.


alexsapps:
@philipp-spiess just letting you know i sent this PR. no worries if you want to make any edits or close it. i figure the v3 and v4 docs should generally be separate but perhaps adding this in the last-ditch troubleshooting section wouldn't hurt.

PR #2153, opened 2025-03-15
Fix link to Tailwind CLI tab on Framework Guides page

marshalhayes:
I noticed that the Tailwind CLI link at the bottom of the framework guides page does not work.

This PR fixes that by changing the link from `/docs/installation` to `/docs/installation/tailwind-cli`.


philipp-spiess (approved):
Thanks!

Issue #2157, opened 2025-03-16
Include official tailwindcss plugins in the documentation


chris23lngr:
The new Tailwind v4 documentation contains only a [minimal](https://tailwindcss.com/docs/functions-and-directives#plugin-directive) explanation of how to install Tailwind plugins. 

I think it might be helpful, especially for new users, to at least provide a list of the official plugins with an explanation (one-liner) of what these plugins are needed for - this is completely missing in the new documentation, in the old one there was at least a small mention.

Even if the JS plugins are outdated, they are often still helpful and deserve a mention (at least in my opinion).

PR #2158, opened 2025-03-16
Add official plugins list to @plugin directive section

chris23lngr:
This PR refers to the issue #2157 

As explained in the issue, I have added a list of the official Tailwind CSS plugins with a link and a short explanation (taken from the README of the respective plugin).

Screenshot of the newly added list at [/docs/functions-and-directives#plugin-directive](https://tailwindcss.com/docs/functions-and-directives#plugin-directive):

Bildschirmfoto 2025-03-16 um 14 54 57

### DIFF
--- a/src/docs/functions-and-directives.mdx
+++ b/src/docs/functions-and-directives.mdx
@@ -263,7 +263,11 @@ Use the `@plugin` directive to load a legacy JavaScript-based plugin:
 @plugin "@tailwindcss/typography";
 ```
 
-The `@plugin` directive accepts either a package name or a local path.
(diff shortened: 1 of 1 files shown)


philipp-spiess:
You're right that these are official tailwind plugins but we decided to not propagate them on the docs page for now since modern v4 projects rarely need them (and, specifically the aspect-ratio one, is now backed into Tailwind 4 too). Plugins also require a slower compatibility code path right now that we don't want people to experience by adding these plugins light-heartedly.
Ideally we'll be adding CSS only replacements for typography/forms then we can reconsider but for now I think it's more in our interest to not give these plugins too much exposure! 
(Just so I'm clear: typography and forms are still very much supported, see their respective READMEs for how they work with v4!)

PR #2170, opened 2025-03-25
fix: properly setup rtl truncation to truncate from the end

karimshalapy:
The current behavior of the RTL part in the `WhyTailwindCssSection` is wrong, and I'll describe the problem in the following points.
- The RTL part does not have `dir="rtl"`
The DOM elements are ordered manually instead of leaving the order the same, and depending on the browser to order them according to their direction. So, in the RTL version, the image is the second element, but in the LTR, it's the first.
The `text-right` is used explicitly on the RTL version, although it should have been prefixed with `rtl:` or omitted entirely if the `dir` is set up correctly.
- The missing `dir` causes another issue with the truncation, as the direction of the container is `ltr`, so the truncation happened from the start of the text, not the end. 
How it looks: image
How it should look: image

These issues are in the RTL version on the homepage, and I didn't check any other pages, but the landing page is important to have an accurate representation to appeal to the developers who need RTL support. If a developer opens the dev tools currently and checks the implementation, they might move away from this solution, although it's capable, just because they saw a wrong implementation.

### DIFF
--- a/src/components/home/why-tailwind-css-section.tsx
+++ b/src/components/home/why-tailwind-css-section.tsx
@@ -634,7 +634,7 @@ export default function WhyTailwindCssSection() {
                       <div key={idx} className="flex items-center justify-start gap-4 p-6">
                         <div className="shrink-0">
                           <img
-                            alt=""
+                            alt={`${user.name} profile picture`}
                             src={user.src}
                             className="size-12 shrink-0 rounded-full bg-gray-950/5 outline -outline-offset-1 outline-gray-950/10 dark:outline-white/10"
                           />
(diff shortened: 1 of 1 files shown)


src/components/home/why-tailwind-css-section.tsx
```diff
@@ -634,7 +634,7 @@ export default function WhyTailwindCssSection() {
                       <div key={idx} className="flex items-center justify-start gap-4 p-6">
                         <div className="shrink-0">
                           <img
-                            alt=""
+                            alt={`${user.name} profile picture`}
```
philipp-spiess, line 637:
```suggestion
                            alt=""
```

We want to keep this as an empty alt text since the name is already included in the DOM anyway and for a person using a screen reader there's no added information in having them know there's a profile picture here (otherwise it'd read `${name}, ${name} profile picture` for every name

src/components/home/why-tailwind-css-section.tsx
```diff
@@ -654,18 +654,19 @@ export default function WhyTailwindCssSection() {
...
-                        <div className="flex flex-col truncate text-right">
-                          <span className="text-sm/6 font-medium text-gray-950 dark:text-white">{user.name}</span>
-                          <span className="truncate text-sm/6 text-gray-500 dark:text-gray-400">{user.role}</span>
-                        </div>
-                        <div className="shrink-0">
+                      <div key={idx} className="flex items-center justify-start gap-4 p-6" dir="rtl">
+                         <div className="shrink-0">
                           <img
-                            alt=""
+                            alt={`صورة شخصية لـ ${user.name}`}
```
philipp-spiess, line 660:
```suggestion
                            alt=""
```

philipp-spiess (approved):
This is great thank you for the fix! I'm going to revert the `alt` changes though since the user name is already included in the DOM for screen readers.

PR #2181, opened 2025-03-28
Document `@source not` and `@source inline`

danhollick:


src/docs/detecting-classes-in-source-files.mdx
```diff
@@ -181,3 +195,120 @@ Use `source(none)` to completely disable automatic source detection if you want
 ```
 
 This can be useful in projects that have multiple Tailwind stylesheets where you want to make sure each one only includes the classes each stylesheet needs.
+
+## Safelisting specific utilities
+
+If you need to make sure Tailwind generates certain class names that don’t exist in your content files, you can force them to be inlined by using `@source inline()`:
```
reinink, line 201: Slight reordering to avoid having a comment as the last item in the paragraph. Also changed "inlined" to "generated", because that feels more natural to me when talking about how Tailwind works.

```suggestion
If you need to make sure Tailwind generates certain class names that don’t exist in your content files, use `@source inline()` to force them to be generated:
```

src/docs/detecting-classes-in-source-files.mdx
```diff
@@ -181,3 +195,120 @@ Use `source(none)` to completely disable automatic source detection if you want
...
+If you need to avoid specific classes from being generated you can use `@source not inline()`:
+
+```css
+/* [!code filename:CSS] */
+@import "tailwindcss";
+/* [!code highlight:2] */
+@source not inline("{hover:,focus:,}bg-red-{50,{100..900..100},950}");
+```
+
+This will specifically exclude the shades of red from 50 to 950, along with their hover and focus variants, from being generated even if they are detected in your source files.
```
reinink, line 314: Tightened up a bit.

```suggestion
This will explicitly exclude the red background utilities, along with their hover and focus variants, from being generated.
```

reinink (changes requested):
Back to you, Dan! 🤙

src/docs/detecting-classes-in-source-files.mdx
```diff
@@ -167,6 +168,19 @@ To set the base path for source detection explicitly, use the `source()` functio
...
 
+### Ignoring specific paths
+
+Use `@source not` to ignore specific paths when scanning for class names:
+
+```css
+/* [!code filename:CSS] */
+@import "tailwindcss";
+/* [!code highlight:2] */
+@source not "src/components/legacy";
```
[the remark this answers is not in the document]
RobinMalfait: I personally like starting with `./` or `../`, gets rid of the confusion where files/folders are relative to.

src/docs/detecting-classes-in-source-files.mdx
```diff
@@ -181,3 +195,120 @@ Use `source(none)` to completely disable automatic source detection if you want
...
+@media (hover: hover) {
+  .hover\\:underline:hover {
+    text-decoration: underline;
+  }
+}
+@media (focus: focus) {
+  .focus\\:underline:focus {
+    text-decoration: underline;
+  }
+}
```
RobinMalfait, line 246: In CSS these are only escaped once:
```suggestion
/* [!code filename:Generated CSS] */
.underline {
  text-decoration: underline;
}
@media (hover: hover) {
  .hover\:underline:hover {
    text-decoration: underline;
  }
}
@media (focus: focus) {
  .focus\:underline:focus {
    text-decoration: underline;
  }
}
```

src/docs/detecting-classes-in-source-files.mdx
```diff
@@ -181,3 +195,120 @@ Use `source(none)` to completely disable automatic source detection if you want
...
+@media (hover: hover) {
+  .hover\\:bg-red-50:hover {
+    background-color: var(--color-red-50);
+  }
+
+  /* ... */
+
+  .hover\\:bg-red-950:hover {
+    background-color: var(--color-red-950);
+  }
```
RobinMalfait, line 295:
```suggestion
@media (hover: hover) {
  .hover\:bg-red-50:hover {
    background-color: var(--color-red-50);
  }

  /* ... */

  .hover\:bg-red-950:hover {
    background-color: var(--color-red-950);
  }
```

src/docs/detecting-classes-in-source-files.mdx
```diff
@@ -181,3 +195,120 @@ Use `source(none)` to completely disable automatic source detection if you want
...
+}
+```
+
+</CodeExampleStack>
+
+This generates shades of red from 100 to 900 in increments of 100, along with the first and last shades of 50 and 950. It also adds the `hover:` variant for each of those classes.
```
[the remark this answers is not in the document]
danhollick: @reinink Tending towards keeping the "in increments of 100" because it explains the third argument in the input.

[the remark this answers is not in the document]
reinink: @danhollick Yup all good, if you think it's worth having, keep it 👍

PR #2196, opened 2025-04-05
Fix incorrect text-shadow color name in API table

dcastil:
`shadow-<color>` is used for box-shadow, the right name is `text-shadow-<color>` according to https://github.com/tailwindlabs/tailwindcss/pull/17389.


philipp-spiess (approved):
Oof yep, thank you!

philipp-spiess (approved):

PR #2203, opened 2025-04-11
Fix `justify-self-center` example label

### LINKED ISSUE #2200: wrong description for justify-self in example
https://tailwindcss.com/docs/justify-self#center

in the code `justify-self-center` is used and in the example `justify-center-end` instead.
Image

wongjn:
Fixes #2200


philipp-spiess (approved):
Hehe thank you sir!

PR #2210, opened 2025-04-16
Fix `text-pretty` description

holloway:
Current `text-pretty` description says,

> Use the `text-pretty` utility to prevent orphans (a single word on its own line) at the end of a text block:

See [this recent blog post from Webkit about `text-wrap: pretty;`](https://webkit.org/blog/16547/better-typography-with-text-wrap-pretty/) and in particular, 

> Because of the way Chrome’s implementation of pretty has been taught, a lot of web developers expect this value is only supposed to prevent short last lines. But that was never the intention. In fact, the CSS Working Group defined a different value for such a purpose. It was [just renamed](https://github.com/w3c/csswg-drafts/issues/11283) last week to text-wrap: avoid-short-last-lines.

So it seems an updated description is needed and this PR has a first attempt.

### DIFF
--- a/src/docs/text-wrap.mdx
+++ b/src/docs/text-wrap.mdx
@@ -111,7 +111,7 @@ For performance reasons browsers limit text balancing to blocks that are ~6 line
 
 ### Pretty text wrapping
 
-Use the `text-pretty` utility to prevent orphans (a single word on its own line) at the end of a text block:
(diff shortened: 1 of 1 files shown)


holloway:
fyi @danhollick who I think wrote [this page](https://github.com/tailwindlabs/tailwindcss.com/pull/1745/commits/4394c8d9e3251e136897d4305fe0e6ef743796f1)?

Issue #2213, opened 2025-04-18
Outdated standalone reference in the v4 CLI installation steps


rozsazoltan:
* https://tailwindcss.com/docs/installation/tailwind-cli

> The simplest and fastest way to get up and running with Tailwind CSS from scratch is with the Tailwind CLI tool. The CLI is also available as a [standalone executable](https://tailwindcss.com/blog/standalone-cli) if you want to use it without installing Node.js.

thecrypticace:
I updated the link to point to the GitHub releases page a bit ago. I think I can close this. The steps are otherwise the same as the normal CLI except you use the binary instead of `npx @tailwindcss/cli`

Issue #2212, opened 2025-04-18
Determine if the content is potentially outdated based on the date of the blog posts


rozsazoltan:
It's misleading when the version number indicates that I'm reading a 4.1 page, but the blog content from 2021 is actually related to version 3. I don't have much of an issue with this, but sometimes I get caught off guard by it. I think it could be easily implemented with a warning message, possibly with a link pointing to the current content.

* https://github.com/tailwindlabs/tailwindcss/discussions/15791#discussioncomment-12876794

<img src="https://github.com/user-attachments/assets/8d2f33e7-ba0d-4b50-b552-c31584eb84c8" width="670">

I'm not sure if it's necessary, but I'll leave it as an idea.

rozsazoltan:
This solution applies exclusively to the Standalone CLI: #2214

philipp-spiess:
I don't think this is a very big issue -- When you're reading a blog post it does not necessarily feel like it has to be updated to the new API, that's why there's a date on the very top. Blog posts are written to be read closer to when they are published and I don't think I have the expectations that the examples are evergreen. 

The Standalone one is a bit special since it is linked from within the docs so I agree that https://github.com/tailwindlabs/tailwindcss.com/issues/2213 is something we want to fix 👍

PR #2214, opened 2025-04-18
v4-compatible standalone blog post following the old v3 one

rozsazoltan:
Fixes: #2213 and partially #2212

* v3-related: `/blog/standalone-cli`
* v4-related: `/blog/2025-04-18-standalone-cli-for-v4`

I didn't make many changes to the original standalone CLI post. I updated the installation steps, removed the init process, and added a note that the `tailwind.config.js` file is no longer necessary. I also included a link to the CSS-first configuration guide.

Additionally, I made a small update to the old post to reference the new one.

I added myself as the author, but I'm not sure what the usual convention is; feel free to change it.


philipp-spiess:
Hey! Thanks but I feel like a blog post is not a good spot to document the Standalone CLI, we should definitely have something in the docs that we keep updated, though.

Adding a blog post would indicate that something _new_ has happened just now which I don't really think is the case beyond the v4 release that changed the CLI. 

Maybe instead there could be a version of https://tailwindcss.com/docs/installation/tailwind-cli that talks about how to install the standalone build (rendered like e.g. https://tailwindcss.com/docs/installation/framework-guides/laravel/vite but not necessarily linked from anywhere but the link in https://tailwindcss.com/docs/installation/tailwind-cli)?

PR #2218, opened 2025-04-21
Demo cursor classes on the rows

jmg-duarte:
Brings back the behaviour of the v3 docs, where hovering the rows would show the cursor.

I recognize that the way I did it is a bit of a dirty hack, feedback is most welcome!

### DIFF
--- a/src/components/cursor-table.tsx
+++ b/src/components/cursor-table.tsx
@@ -1,0 +1,148 @@
+"use client";
+
+import clsx from "clsx";
+import { useEffect, useRef, useState } from "react";
+
+function highlight(input: string) {
+  return (
+    <code>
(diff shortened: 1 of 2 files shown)


philipp-spiess:
Appreciate the change here but I don't think the added cursor preview here is worth forking the whole API table for at the moment, so not going to accept this right now, thanks though!

jmg-duarte:
> Appreciate the change here but I don't think the added cursor preview here is worth forking the whole API table for at the moment, so not going to accept this right now, thanks though!

@philipp-spiess I can agree with that but I'm keen on changing to another approach with a little guidance / some idea of what would be acceptable. I was just looking to have this since I had to go back to v3 to check the previews

PR #2219, opened 2025-04-21
add docs for border-width and opacity theme namespaces

phaux:


philipp-spiess:
Thanks! Agree that we need do document these properties better -- we consider them somehow a backward compatibility feature since it worked with v3 so we made it work with v4, however it's not something that we found being used much in practice.

If/when we document those, we'd need to be exhaustive though and include all of the currently-undocumented namespaces, only adding two of them here sets expectations that these are equally-important as the other namespaces right now which we don't think is true. Going to close this PR for now but yeah I do agree we need them _somewhere_ in the docs.

kachkaev:
Out of curiosity, why is `--border-width-*` deprecated? Instead of `--opacity-*`, we can use `bg-red-500/[opacity]` which makes sense. When it comes to border width though, I don't see an alternative. For example, if I need `border-16` or `border-hairline` in my theme, the only possible way in v4 seems to be setting `--border-width-*`. If that's not true, what is the right alternative?

PR #2222, opened 2025-04-27
[added a copy button]

sayanChaterjee:
I have added a copy button which will enable the users to copy the code snippets of the steps of installation. Manually copying each and every installation code is tedious.

### DIFF
--- a/src/components/code-example.tsx
+++ b/src/components/code-example.tsx
@@ -190,11 +190,18 @@ export function RawHighlightedCode({
   return <div className={className} dangerouslySetInnerHTML={{ __html: code }} />;
 }
 
+function cleanCodeForCopy(code:string) {
+  return code
+    .split('\n')
+    .filter(line => !line.includes('[!code highlight'))
+    .join('\n');
+}
+
 function CodeExampleFilename({ filename, code }: { filename: string; code?: string }) {
   return (
     <div className="flex justify-between px-3 pt-0.5 pb-1.5 text-xs/5 text-gray-400 dark:text-white/50">
       {filename}
-      {code && <CopyButton code={code} />}
+      {code && <CopyButton code={cleanCodeForCopy(code)} />}
     </div>
   );
 }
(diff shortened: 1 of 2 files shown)


rozsazoltan:
It doesn't work perfectly for me, the code keeps the highlights as comments. I think for any copy-paste, such comments shouldn't be kept.

* https://tailwindcss-1qh7qshdg-tailwindlabs.vercel.app/docs/installation/using-vite

```js
import { defineConfig } from 'vite'
// [!code highlight:2]
import tailwindcss from '@tailwindcss/vite'

export default defineConfig({
  plugins: [
    // [!code highlight:2]
    tailwindcss(),
  ],
})
```

sayanChaterjee:
Thanks for addressing the code highlighting issue. I've verified that the bug has been fixed. The copy functionality now correctly strips out highlight directives before copying code snippets to the clipboard, ensuring users receive clean, ready-to-use code without unnecessary comments.

sayanChaterjee:
@rozsazoltan Hi there! Just checking in on this PR one more time - I've resolved the issue you mentioned and pushed the fix in my most recent commit. Would greatly appreciate it if you could take another glance when you have the opportunity. This feature should simplify things significantly for users once it's live! If anything else needs adjusted, let me know.

src/components/code-example.tsx
```diff
@@ -189,8 +190,20 @@ export function RawHighlightedCode({
   return <div className={className} dangerouslySetInnerHTML={{ __html: code }} />;
 }
 
-function CodeExampleFilename({ filename }: { filename: string }) {
-  return <div className="px-3 pt-0.5 pb-1.5 text-xs/5 text-gray-400 dark:text-white/50">{filename}</div>;
+function cleanCodeForCopy(code:string) {
+  return code
+    .split('\n')
+    .filter(line => !line.includes('[!code highlight'))
```
philipp-spiess, line 196: Hey! Sorry I haven't had time to properly review this yet but this line does not look like it's generic enough to handle all the code highlighter annotations. Some of them appear inline (so at the end of a line containing code). 

This also doesn't handle diff views at all: https://tailwindcss-com-git-fork-sayanchaterjee-fea-b7f6d8-tailwindlabs.vercel.app/docs/upgrade-guide#using-postcss

Maybe a better way is to either use the output from shiki somehow or rely on the DOM state to get the _visible code_? 

I'm not sure what I'd expect to happen there for the diff example though. Maybe it's fine because in v3 this also didn't do anything special, see app.js [here](https://v3.tailwindcss.com/docs/guides/phoenix).

PR #2224, opened 2025-04-28
feat/fix copy button

sayanChaterjee:
I have added a copy button which will enable the users to copy the code snippets of the steps of installation. Manually copying each and every installation code is tedious. It also resolves the code highlighting issue. I've verified that the bug has been fixed. The copy functionality now correctly strips out highlight directives before copying code snippets to the clipboard, ensuring users receive clean, ready-to-use code without unnecessary comments.


sayanChaterjee:
> Duplicated: #2222

Yeah. I have fixed the issue which occurred in #2222

PR #2231, opened 2025-05-03
Fix example syntax error

wongjn:
Fix unbalanced parentheses in an example, as spotted by tailwindlabs/tailwindcss#17857.


RobinMalfait (approved):
Oopsie, nice one thanks!

Issue #2240, opened 2025-05-12
Some <h3> elements do not contain an anchor


rozsazoltan:
Typically, Markdown headings like `#`, `##`, and `###` are used, and their compiled HTML versions contain the appropriate link references.

**./src/docs/theme.mdx**
```md
#### Relationship to variants
```

```html
<h4 id="relationship-to-variants">
  <a href="#relationship-to-variants" class="anchor">Relationship to variants</a>
</h4>
```

However, in the case of custom `h1`, `h2`, `h3`, etc. elements, this property is not present by default.

**./src/docs/functions-and-directives.mdx**
```html
<h3 id="import-directive">@import</h3>
```

```html
<h3 id="import-directive">
  @import
</h3>
```

It's a bit confusing that there is an ID pointing to the heading, but it's harder to copy because clicking on the heading doesn't navigate to itself.

PR #2241, opened 2025-05-12
Added anchors to custom headings

rozsazoltan:
Fixes: #2240 

In the spirit of reusability, I declared a new component called `Heading`. You need to specify the level (between 1 and 6), the ID (which is mandatory), and the content of the element. With this in place, I also declared `H2` through `H6` elements. Thus, instead of:

```html
<h3 id="import-directive">@import</h3>
```
* [Preview](https://tailwindcss.com/docs/functions-and-directives#import-directive)

you can use:
```html
<H3 id="import-directive">@import</H3>
```
* [Preview](https://tailwindcss-com-git-fork-rozsazoltan-fix-he-f001db-tailwindlabs.vercel.app/docs/functions-and-directives#import-directive)

what will be:
```html
<h3 id="import-directive">
  <a href="#import-directive" class="anchor">@import</a>
</h3>
```


philipp-spiess:
@rozsazoltan There's another place right now where we just add that `<a>` tag manually: https://github.com/tailwindlabs/tailwindcss.com/blob/main/src/docs/responsive-design.mdx?plain=1#L476-L478 I think we should just go ahead and do that instead of adding a heading component. It also seems that the anchor class isn't really necessary: https://tailwindcss.com/docs/responsive-design#using-arbitrary-container-query-values

rozsazoltan:
Ah, indeed, that slipped my attention. Thanks for the update!

PR #2245, opened 2025-05-12
Fix broken link in Tailwind v3 docs

ryanjbonnell:


src/pages/docs/configuration.mdx
```diff
@@ -43,7 +43,7 @@ module.exports = {
 }
 ```
 
-Every section of the config file is optional, so you only have to specify what you'd like to change. Any missing sections will fall back to Tailwind's [default configuration](https://github.com/tailwindlabs/tailwindcss/blob/master/stubs/config.full.js).
+Every section of the config file is optional, so you only have to specify what you'd like to change. Any missing sections will fall back to Tailwind's [default configuration](https://github.com/tailwindlabs/tailwindcss/blob/v3.4.17/stubs/config.full.js).
```
rozsazoltan, line 46: Instead of referring to the latest v3 tag, I would refer to the v3 branch.

* https://github.com/tailwindlabs/tailwindcss/blob/v3/stubs/config.full.js

philipp-spiess (approved):

philipp-spiess:
Thanks!

PR #2178, opened 2025-03-28
Document specific theme keys

crswll:
This PR adds a small section to the `CustomizingYourThemeColors` component to document how to add colors for a certain property specifically.

PR #2247, opened 2025-05-13
Added docs for utility-scoped custom colors

rozsazoltan:
At first, I felt it was missing from the Theme documentation page, but I ended up adding it to the Colors section instead, as the first subsection under customization.

```css
@theme {
  --background-color-midnight: #121063; /* for only bg-midnight */
  --text-color-tahiti: #3ab7bf;         /* for only text-tahiti */
  --border-color-bermuda: #78dcca;      /* for only border-bermuda */
}
```


rozsazoltan:
Duplicated: #2178

PR #2252, opened 2025-05-15
fix: fix inconsistent min-width fraction example

resetsix:
The example in min-width.mdx shows min-w-1/3 in the UI but uses min-w-3/4 in the HTML example.  
This commit updates the UI example to correctly display min-w-3/4 to match the HTML code.


RobinMalfait (approved):
Thanks!

PR #2258, opened 2025-05-20
fix background-position.mdx

ArnaudBan:
The background position where wrong : `bg-top-left`, `bg-top-right`, `bg-bottom-left` and `bg-bottom-right` don't existe


adamwathan:
Hey! They do actually, just make sure you're on the latest version:

https://github.com/tailwindlabs/tailwindcss/pull/17378

PR #2259, opened 2025-05-23
docs(nextjs): clarify globals.css placement for App Router

uliba3:
Enhance Next.js guide by adding a step to import global.css and introduce a new tsx utility for code examples.

### DIFF
--- a/src/app/(docs)/docs/installation/framework-guides/nextjs.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/nextjs.tsx
@@ -1,4 +1,4 @@
-import { css, js, shell, Page, Step, Tile, tsx } from "./utils";
+import { css, js, shell, Page, Step, Tile } from "./utils";
 import Logo from "@/docs/img/guides/nextjs.react.svg";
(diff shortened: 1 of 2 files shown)


src/app/(docs)/docs/installation/framework-guides/nextjs.tsx
```diff
@@ -85,6 +85,21 @@ export let steps: Step[] = [
       `,
     },
   },
+  {
+    title: "Import global.css",
+    body: (
+      <p>
+        Import the global.css file in the <code>./src/app/layout.tsx</code> file.
```
rozsazoltan, line 92: If I follow the documentation, then this is an unnecessary step.
* https://github.com/vercel/next.js/blob/canary/packages/create-next-app/templates/app/ts/app/layout.tsx#L3

uliba3: I ran into this myself and spent hours debugging because I had removed the default globals.css import from layout.tsx. Thought it might help others — but I see now the default template covers it and the other guides don’t mention this either. Thanks for the review!

PR #2260, opened 2025-05-23
Update Tailwind CSS integration example for Laravel

AhmedAlaa4611:
Description
---
This PR updates the documentation to reflect the actual default integration of Tailwind CSS in Laravel as shown in the official [Laravel repository](https://github.com/laravel/laravel/blob/12.x/resources/css/app.css).

Reason
--
The current Tailwind documentation for Laravel does not accurately reflect how Tailwind is integrated by default in the latest Laravel releases. This update ensures that new users following the guide get a setup experience that matches the actual Laravel boilerplate.

### DIFF
--- a/src/app/(docs)/docs/installation/framework-guides/laravel.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/laravel.tsx
@@ -130,15 +130,19 @@ export let steps: Step[] = [
     body: (
       <p>
         Add an <code>@import</code> to <code>./resources/css/app.css</code> that imports Tailwind CSS. Additionally,
-        tell Tailwind CSS to scan your <code>resources/views</code> directory for utilities.
+        tell Tailwind CSS to scan some directories for utilities.
       </p>
     ),
     code: {
       name: "app.css",
       lang: "css",
       code: css`
-        @import "tailwindcss";
-        @source "../views";
+        @import 'tailwindcss';
+
+        @source '../../vendor/laravel/framework/src/Illuminate/Pagination/resources/views/*.blade.php';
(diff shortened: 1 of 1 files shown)


rozsazoltan:
> ```css
> @source '../../vendor/laravel/framework/src/Illuminate/Pagination/resources/views/*.blade.php';
> ```

The link pointing to the vendor folder makes sense; but if you're not using pagination - like me - it seems like an unnecessary guide, especially since the Laravel Docs specifically cover it in the Pagination section.
* https://github.com/laravel/docs/pull/10238
* https://laravel.com/docs/12.x/pagination#tailwind

> ```css
> @source '../**/*.blade.php';
> @source '../**/*.js';
> ```

When building, I'd rather exclude `storage/framework/views` than inject it - but that's a matter of personal relevance. If adding the views is truly important, it's justified by the `.gitignore`, as it does indeed exclude them, which causes automatic source detection to miss them.
* https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned
* https://github.com/laravel/laravel/blob/12.x/storage/framework/.gitignore

> ```css
> @source '../../storage/framework/views/*.php';
> ```

I didn't see much point in the other two either, since automatic source detection should handle all of them anyway. I see the point of these only if the source is disabled or specifically configured during `@import`. In that case, completing the paths is indeed useful. However, the docs recommend not doing that and instead using `@import "tailwindcss";` on its own.

thecrypticace:
Alright, I got some clarification on this. There *are* some reasons to include compiled views in the app related to packages that register and use views but aren't directly used by the app.

PR #2261, opened 2025-05-25
Fixed the app not-found height to hide the scrollbar

rozsazoltan:
The parent element has a `pt-14.25`, which needs to be subtracted from `100vh`. The parent also has a `1px` border on the top and bottom, which should also be subtracted. Additionally, I would use `dvh` instead of `vh`.

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/src/app/not-found.tsx
+++ b/src/app/not-found.tsx
@@ -16,7 +16,7 @@ export default function NotFoundPage() {
 
         {/* Main content area */}
         <div className="text-gray-950 dark:text-white">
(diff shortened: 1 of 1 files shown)


philipp-spiess (changes requested):
I _think_ (from trying in DevTools in Chrome only) it would also work if you add a `h-[100dvh]` on to the outer most div in this component (`<div className="max-w-screen overflow-x-hidden">`) and then use a `h-full` instead of this `h-[calc(100dvh-var(--spacing)*14.25-2px)]`:  Screenshot 2025-05-26 at 11 01 30

PR #2265, opened 2025-05-30
Enhance the MDX setup

remcohaszing:
This fixes a couple of things.
- The argument was removed from `useMDXComponents()`. In reality this function doesn’t receive any arguments.
- Components overrides are defined outside of `useMDXComponents()`. `useMDXComponents()` is called during render. Defining React components during render causes unnecessary re-renders and loss of state.
- The `**/*.mdx` glob is included in the `tsconfig.json` include patterns. This makes `.mdx` files part of the same TypeScript program in the editor. This leads to a better editor experience and reduced memory and CPU usage.
- This augments the `mdx/types` module to define the `JSX` namespace. React 18 defined the global `JSX` namespace. React 19 no longer does this. This leads to type errors in `mdx/types` resulting unresolved `any` types. This augmentation makes the MDX types understand React.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,6 +22,6 @@
       "@/*": ["./src/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "include": ["next-env.d.ts", "**/*.mdx", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
   "exclude": ["node_modules"]
 }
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -22,6 +22,6 @@
       "@/*": ["./src/*"]
     }
   },
```
remcohaszing, line 24: You can also add the following to change the subtle TypeScript MDX editor hints into red squiggly type errors inside MDX files:

```suggestion
  },
  "mdx": {
    "checkMdx": true
  },
```

remcohaszing:
Yes. I’ve tried convincing them several times, but sadly they won’t believe me. It’s quite frustrating.

If you need credibility: I’m an MDX maintainer. If you need proof: Add `console.log(arguments)` inside `useMDXComponents()` or look at the MDX compile output.

thecrypticace:
Oh… 😅 

Will take a look at this sometime this weekend probably. Thanks!

thecrypticace (approved):

PR #2268, opened 2025-06-07
Add attribute alt to images in landing page

GeorgiosDrivas:
After running a lighthouse test on the website, I discovered some empty alt attributes on some images.
For readability and accessibility issues I filled those alt attributes based on the images.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6942 @@
+{
+  "name": "v4.tailwindcss.com",
+  "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


philipp-spiess:
Hey man! Sorry for the super late response but this is all on purpose since the images do not add information that is not already available as text content, see: https://www.w3.org/WAI/tutorials/images/decorative/ so going to close this but thanks for thinking about accessibility!

PR #2278, opened 2025-06-15
Fix/feature copy to clipboard #2271

Md-Mahir-Asef:
This pull request adds Copy button to all code snippets in the tailwindcss.com documentation. Here are some screen shorts.
### **Before**:
Screenshot 2025-06-14 203340
Screenshot 2025-06-14 203401
Screenshot 2025-06-15 101209

### **After**:
Screenshot 2025-06-14 203044
Screenshot 2025-06-14 203117
Screenshot 2025-06-14 203218

PR #2279, opened 2025-06-16
Added copy button

### LINKED ISSUE #2271: [Feat] : Add "Copy to Clipboard" button on code snippets in the documentation
# ✨ Feature Request: Add "Copy to Clipboard" Button for Code Snippets

## Description

Currently, on the [Tailwind CSS documentation](https://tailwindcss.com/docs/installation/using-vite), the code snippets shown in the right-hand editor-style panels (e.g., under the "Getting Started" section) do **not include a "Copy to Clipboard"** button. 

This requires users to manually highlight the code and use `Ctrl+C` or long-press on mobile to copy it, which adds a small friction point — especially when working with multi-line commands or configuration code.

---

## Steps to Reproduce

1. Navigate to: https://tailwindcss.com/docs/installation
2. Locate any code block within the documentation (e.g., Terminal commands, `vite.config.ts`, CSS import lines).
3. Observe that the code is shown inside a stylized editor box.
4. Try copying the code — it must be manually selected and copied.
5. There is no dedicated "Copy" button available in the UI.

---

## Screenshots

Image

Image

---

## Suggested Enhancement

Add a **"Copy to Clipboard"** button to the top-right corner of each code snippet block across the documentation.

### Desired Behavior:

- When clicked, the button should copy the full content of the associated code block to the clipboard.
- A brief visual confirmation (e.g., tooltip or toast saying _"Copied!"_) should appear upon success.
- The feature should be:
  - **Accessible** (keyboard support and ARIA labels for screen readers),
  - **Mobile-friendly**, and
  - **Consistent across all documentation pages**.

---

## Benefits

- Improves usability and developer experience.
- Saves time during setup and onboarding.
- Makes the documentation more interactive and accessible.

---

Thanks for considering this suggestion! Tailwind CSS has been a fantastic tool in my workflow, and I’d love to see its already-great documentation made even better with this small improvement.

rozsazoltan:
Closes #2271

I know there are already a few active PRs (#2222, #2278), but each of them got stuck at some point for various reasons and tries to filter out Shiki comments using different logic.

There are several issues to address:
- we need to be compatible with different comment syntaxes across various languages  
- shiki comments can appear not only on new lines but also inline  
- in the case of inline comments, we don’t want to remove the entire line  
- for diff code, deleted lines aren't needed - after all, that's the point of diff examples  
- it needs to handle the removal of all Shiki comments in a general way, without listing every possible Shiki comment syntax explicitly

For my part, I'm currently using the code shared in this PR - it seems to be working so far:
- the copy button appears even without a code example header  
- it handles diffs correctly  
- it processes `highlight`, `word`, and other expressions properly  
- it works with any kind of code example (bash, css, html, js, ...)

I'm not convinced that this is the best solution to the problem - that's why I hesitated to share it earlier.

### DIFF
--- a/src/components/code-example.tsx
+++ b/src/components/code-example.tsx
@@ -8,6 +8,7 @@ import { clsx } from "clsx";
 import dedent from "dedent";
 import { createHighlighter } from "shiki";
 import theme from "./syntax-highlighter/theme.json";
+import { CopyButton } from "./copy-button";
 
 import { highlightClasses } from "./highlight-classes";
 import atApplyInjection from "./syntax-highlighter/at-apply.json";
@@ -97,8 +98,12 @@ export async function CodeExample({
   className?: string;
 }) {
   return (
-    <CodeExampleWrapper className={className}>
+    <CodeExampleWrapper className={clsx('relative', className)}>
       {filename ? <CodeExampleFilename filename={filename} /> : null}
+      <CopyButton
+        className="absolute right-4 top-2 z-10 text-stone-400"
+        value={example.code}
+      />
       <HighlightedCode example={example} />
     </CodeExampleWrapper>
   );
--- a/src/components/code-example.tsx
+++ b/src/components/code-example.tsx
@@ -38,6 +38,55 @@ export function css(strings: TemplateStringsArray, ...args: any[]) {
   return { lang: "css", code: dedent(strings, ...args) };
 }
 
+export function unshiki(code: string): string {
+  const lines = code.split("\n");
+  const result: string[] = [];
+  let skip = 0;
+
+  const commentRegex = /\/\*.*?\*\/|\/\/.*|<!--.*?-->|#.*/g;
+  const codeTagRegex = /\[!code\s+([^\]]+)\]/;
+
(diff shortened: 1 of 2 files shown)


src/components/code-example.tsx
```diff
@@ -38,6 +38,55 @@ export function css(strings: TemplateStringsArray, ...args: any[]) {
...
+        }
+
+        // remove comment if it's not a remove directive
+        line = line.slice(0, c.index) + line.slice(c.index! + c[0].length);
+      }
+    }
+
+    // add line if not removed and line is not empty or has no comments
+    if (!removed && (comments.length === 0 || line.trim() !== "")) {
+      result.push(line);
```
rozsazoltan, line 84: After that, we add to the result those lines that don't need to be removed (special case) and either had no comments or had comments that were removed, leaving the line empty.

thecrypticace:
little update: We're gonna design a custom icon for the copy button and *maybe* switch to a "copied" tooltip kinda like we had in the v3 docs but not 100% sure yet. Likely will get things updated and merged sometime next week.

RobinMalfait (approved):

PR #2282, opened 2025-06-19
Microsoft Copilot showcase

marshalhayes:
I was looking at Microsoft Copilot's source code & noticed they were using Tailwind. 

This PR adds Microsoft Copilot as a showcase.

I struggled with the videos quite a bit, so let me know if I need to make any improvements there.


philipp-spiess:
Hey! Thanks for the PR but not going to add this right now. We've started to prefer showcasing more sponsor websites for now. Might be adding Microsoft Copilot at a later point though.

PR #2285, opened 2025-06-20
Fix misleading text in "Hover, focus, and other states" page

benface:
The generated child selector has the same specificity as the regular one, so the reason it can't be overridden is just the CSS order.

### DIFF
--- a/src/docs/hover-focus-and-other-states.mdx
+++ b/src/docs/hover-focus-and-other-states.mdx
@@ -2757,7 +2757,7 @@ While it's generally preferable to put utility classes directly on child element
 
 It's important to note that overriding a style with a utility directly on the child itself won't work since children rules are generated after the regular ones and they have the same specificity:
 
-<TipBad>{<>Won't work, children can't override styles defined on their parent.</>}</TipBad>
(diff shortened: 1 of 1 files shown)


src/docs/hover-focus-and-other-states.mdx
```diff
@@ -2755,9 +2755,9 @@ While it's generally preferable to put utility classes directly on child element
 
 </Figure>
 
-It's important to note that overriding a style with a utility directly on the child itself won't work due to the specificity of the generated child selector:
+It's important to note that overriding a style with a utility directly on the child itself won't work since children rules are generated after the regular ones and they have the same specificity:
 
-<TipBad>{<>Won't work, children can't override their own styling.</>}</TipBad>
+<TipBad>{<>Won't work, children can't override styles defined on their parent.</>}</TipBad>
```
thecrypticace, line 2760: `their own styling` is referencing the fact that `*` places the style on the element itself and not on the parent element. I can see how this is confusing though. 🤔

thecrypticace:
```suggestion
<TipBad>{<>Won't work, children can't override styles given to them by the parent.</>}</TipBad>
```

PR #2289, opened 2025-06-22
Improve Negative Values example

wongjn:
`<length>` × `<percentage>` or `<length>` × `<length>` in `calc()` is invalid. This makes the example misleading. I've rewritten the example utility with valid examples.

### DIFF
--- a/src/docs/adding-custom-styles.mdx
+++ b/src/docs/adding-custom-styles.mdx
@@ -548,12 +548,14 @@ To support negative values, register separate positive and negative utilities in
 ```css
 /* [!code filename:CSS] */
 @utility inset-* {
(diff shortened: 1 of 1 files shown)


src/docs/adding-custom-styles.mdx
```diff
@@ -548,12 +548,14 @@ To support negative values, register separate positive and negative utilities in
 ```css
 /* [!code filename:CSS] */
 @utility inset-* {
-  inset: calc(var(--spacing) * --value([percentage], [length]));
+  inset: calc(var(--spacing) * --value(integer));
```
thecrypticace, line 551:
```suggestion
  inset: --spacing(--value(integer));
```

thecrypticace (approved):

PR #2306, opened 2025-07-16
Prevent infinite compilation loop in Symfony app

Kocal:
Related to https://github.com/symfony/webpack-encore/issues/1374

Users in Symfony apps face issues when running Webpack Encore in watch/dev-server mode, because we generates files in `public/`, but Tailwind detects them as source files, and leads to infinite compilation loop.


thecrypticace (approved):

thecrypticace:
Pretty sure this didn't used to be an issue but it changed at some point — thanks!

PR #2321, opened 2025-08-01
Add details about `hidden` attribute preflight rules

thecrypticace:
In v4 we made a change that makes elements with a hidden attribute `display: none !important`. A consequence of this is that `<div hidden class="block">test</div>` will be invisible in v4 where it was visible in v3.

The intent is that you must *remove* this attribute if the element isn't actually meant to be hidden but we didn't document this in the upgrade guide or the preflight page.

Closes https://github.com/tailwindlabs/tailwindcss/issues/18653

### DIFF
--- a/src/docs/preflight.mdx
+++ b/src/docs/preflight.mdx
@@ -190,7 +190,7 @@ This prevents them from overflowing their containers and makes them responsive b
 }
 ```
 
-This enfoces that elements with a `hidden` attribute stay invisible unless using `hidden="until-found"`. Remove the `hidden` attribute if you want an element to be visible to the user.
+This enforces that elements with a `hidden` attribute stay invisible unless using `hidden="until-found"`. Remove the `hidden` attribute if you want an element to be visible to the user.
 
 ## Extending Preflight
 


src/docs/preflight.mdx
```diff
@@ -181,6 +181,17 @@ This prevents them from overflowing their containers and makes them responsive b
...
+#### Elements with a `hidden` attribute stay hidden
+
+```css
+/* [!code filename:CSS] */
+[hidden]:where(:not([hidden='until-found'])) {
+  display: none !important;
+}
+```
+
+This enfoces that elements with a `hidden` attribute stay invisible unless using `hidden="until-found"`. The should *remove* the `hidden` attribute if you want an element to be visible to the user.
```
maiieul, line 193: The should -> You should

maiieul (commented):
A small typo. Otherwise LGTM!

maiieul:
P.S. imo the issue should be converted to a discussion instead of just closed, as there seem to be strong arguments against this breaking change.

src/docs/preflight.mdx
Wiicolas, line 193: This enfoces -> This enforces

philipp-spiess (approved):

PR #2326, opened 2025-08-05
Fix: background-image.mdx

GOKORURI007:
Custom properties should be wrapped in square brackets rather than parentheses.


thecrypticace:
The parens are correct in v4:

Screenshot 2025-08-11 at 10 28 33

PR #2332, opened 2025-08-12
Fix `justify-stretch` example

### LINKED ISSUE #2327: bad example in docs
https://tailwindcss.com/docs/justify-content

this page contains:

```
<div class="flex justify-stretch ...">
  <div>01</div>
  <div>02</div>
  <div>03</div>
</div>
```

I don't think this example works. flex row does not support justify stretch.

thecrypticace:
Fixes #2327

The example code showed `flex` but `justify-stretch` only works on grid containers. In flex containers this is controlled with the `grow-*` class.


RobinMalfait:
Just for future reference: we couldn't find an example where `justify-stretch` does something _different_ than the default `justify-normal`. The spec talks about "Fill grid area" vs "Use natural size" but not sure what that means in practice...

image

PR #2333, opened 2025-08-13
Fix Incorrect Documentation for Configuring defineConfig in Rsbuild 1.4.13

LingTaixu:
This pull request addresses an issue in the documentation for Rsbuild version 1.4.13, where the usage of `defineConfig` was incorrectly described. The changes update the documentation to reflect the correct configuration approach, ensuring that users can properly configure `defineConfig` without encountering TypeScript errors (e.g., `module` not being a valid property in `RsbuildConfigExport`). The updated documentation clarifies that `module` configurations should be placed under `tools.rspack.module` and provides accurate examples for compatibility with the latest Rsbuild version.


thecrypticace:
This guide is for Rspack not Rsbuild

PR #2348, opened 2025-09-02
fix background-image api map error

eason-woong:
According to the API map above, I thought this part should be `<value>` instead of `<image>`.


RobinMalfait:
Thanks!

PR #2351, opened 2025-09-07
fix: indicate which template to use when installing on vite

datasalaryman:
This change tells the user to use the **Vanilla** vite template to install Tailwind. The installation steps in this section will not work with other Vite examples.

### DIFF
--- a/src/app/(docs)/docs/installation/(tabs)/using-vite/page.tsx
+++ b/src/app/(docs)/docs/installation/(tabs)/using-vite/page.tsx
@@ -22,7 +22,7 @@ const steps: Step[] = [
     body: (
       <p>
         Start by creating a new Vite project if you don’t have one set up already. The most common approach is to use{" "}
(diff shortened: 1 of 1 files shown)


rozsazoltan:
Why? The Tailwind guide works with any template. Why would you force me onto a "vanilla" JavaScript template?

rozsazoltan:
If using the vanilla template were essential for the documentation, that command would be included there:
* https://vite.dev/guide/#scaffolding-your-first-vite-project
```none
npm create vite@latest my-app -- --template vanilla
```

But that's not the case. If you want to use Vite, you should really start with the Vite docs and then move on to TailwindCSS. This create command is really aimed at very beginner developers.

datasalaryman:
Oh, just tested. The error I experienced before is gone. Thanks!

datasalaryman:
Oh, just tested. The error I experienced before is gone. Thanks!

Issue #2357, opened 2025-09-15
`place-content` should change from "block axis" to "both axes"


jhcarl0814:
place-content - Flexbox & Grid - Tailwind CSS (https://tailwindcss.com/docs/place-content ) is using "block axis" in each paragraph, but to be precise the right wording should be "both axes".

PR #2358, opened 2025-09-15
Improve `place-content` wording by referencing the correct axis

RobinMalfait:
This PR fixes an issue where we were only using `block axis` in the `place-content` property docs, but the `place-content` property works on both the inline axis and the block axis.

This PR now references both the inline axis and block axis. This now also talks about the difference between rows and columns because we only talked about rows before.

Fixes: https://github.com/tailwindlabs/tailwindcss.com/issues/2357

### DIFF
--- a/src/docs/place-content.mdx
+++ b/src/docs/place-content.mdx
@@ -154,7 +154,7 @@ Use `place-content-between` to distribute grid items along the inline and block 
 
 ### Space around
 
-Use `place-content-around` to distribute grid items such that there is an equal amount of space around each row on the inline axis and each column on the block axis:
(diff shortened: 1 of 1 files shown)


src/docs/place-content.mdx
```diff
@@ -26,7 +26,7 @@ export const description = "Utilities for controlling how content is justified a
 
 ### Center
 
-Use `place-content-center` to pack items in the center of the block axis:
+Use `place-content-center` to pack items in the center of the inline axis and the block axis:
```
thecrypticace, line 29: It's just a style thing so idc too much but I kinda like this wording:

```suggestion
Use `place-content-center` to pack items in the center of the inline and block axes:
```

thecrypticace (approved):

PR #2345, opened 2025-08-28
Fix tooltip clipping + hover lag in P3 color grid (zero re-renders, 60fps)

Austin1serb:
# Fix tooltip clipping + hover lag in P3 color grid (zero re-renders, 60fps)

Closes #2293. Follow-up to the prior PR ( (#2295) that was closed due to clipping.

## Why
- The previous tooltip caused hover lag due to React re-renders.
- My initial attempt removed re-renders but was clipped by stacking/overflow contexts.
- Target: **no React re-renders**, **no clipping**, **60fps** pointer tracking, **no new deps**.

## What changed
- Use a **single, reused tooltip node** with **`position: fixed`**.
- Compute `top/left` from the hovered swatch via `getBoundingClientRect()` and clamp to the viewport.
- Toggle visibility with a tiny local helper (`createZeroUI`) that mutates a scoped `data-tooltip` attribute—**no React state** on the hover path.
- One read + one write per pointer event; no layout thrash observed.

## Results (local profiling on “/”)
| Metric                     | Before  | After     | Delta     |
|-------------------|--------:|----------:|----------:|
| FPS (Chrome)               | ~15–20  | 60 steady | +40–45    |
| React re-renders          | Many      | 0               | −100%      |

PR #2359, opened 2025-09-17
Fix P3 Colors tooltip lag + clipping 

### LINKED ISSUE #2293: Performance Issue: Significant lag in homepage P3 color grid tooltip due to React rerenders (with measured fix)
## Current problem

The tooltip hover interaction in the **P3 Colors section** of the Tailwind homepage [why-tailwind-css-section.tsx#L457](https://github.com/tailwindlabs/tailwindcss.com/blob/main/src/components/home/why-tailwind-css-section.tsx#L457)

causes noticeable lag due to React rerenders. This impacts animations of surrounding elements and significantly slows scrolling when hovered quickly.

## To Reproduce

Go to [tailwindcss.com](https://tailwindcss.com/) scroll to the P3 colors section, open dev tools or simply looks at the carousel slider above it, or the cubic-bezier animation below it, while hovering over the colors

### Measured performance

* **Current**: <20 FPS, high-frequency React rerenders.
* **Proposed fix**: Stable 60 FPS, **0 rerenders**.

### Visual evidence (recorded on Apple M2, 16GB)

**GIF FPS Preview:**
 
Image

**Example video:**

https://github.com/user-attachments/assets/84ef7996-2dc8-47f4-bc8c-c674907b9b80

### Root cause analysis

The tooltip hover logic relies on React's local state (`useState`) combined with React Portals?, causing extensive, repeated component rerenders and heavy browser lag.

### Proposed solution

Replace React state-based hover handling with a tiny (<500Byte) utility that leverages Tailwinds built-in variants to handle UI updates.
This completely eliminates React rerenders, stabilizes FPS, and improves scroll smoothness without sacrificing current functionality.

Ive already built and tested this approach locally. it has fewer lines of code, identical UX, and dramatic performance improvements.

Happy to submit a PR if you're interested. Let me know of any specific constraints or preferences!

Austin1serb:
# Fix P3 Colors tooltip lag

### Summary

Replace the React-state tooltip in the **homepage P3 color grid** with a single, shared, body-mounted tooltip positioned via `translate3d`. No React state on the hover path -> no commits, no jank, no clipping, no re-renders. 

---

### The Problem
* Chrome/Safari **suppress mouse events ~100ms after scroll**, making `:hover`/mouseenter tooltips appear “stuck”.
* In the current implementation, due to re-renders in chrome, this issue is avoided. but still exists in safari.

---

### New Issues' Found with current Approach 

The tooltip overflows on top of the topbar/header

https://github.com/user-attachments/assets/a54d15cd-b463-4608-b90a-e04a4612bd19

In Safari, the tooltip does not trigger re-renders as mentioned by @thecrypticace in [issue-2293](https://github.com/tailwindlabs/tailwindcss.com/issues/2293#:~:text=Weird%2C%20this%20isn%27t,1).
This causes the tooltip to be stuck in place during scroll:

https://github.com/user-attachments/assets/3d9d5022-087f-4165-a7c0-c8234e48fe94

---

### What I tried (and why it failed)

* **Pure `:hover` / mouseenter**: Breaks during scroll due to Chrome/Safari suppression (see Chrome thread below), on top of that the stacking context in this case was not escape-able in a way that was clean.
* **React Portal w/ state**: Still re-renders on hover -> frame drops. + This does not work in safari (this may be how the devs at headlessui handled the scroll throttling. eg. the re-rendering may have been intentional)
* **`position: fixed`**: Removes clipping but looks wrong while scrolling (review feedback); anchoring drifts from swatch.

Refs: Chrome dev team discussion on scroll behavior and `elementFromPoint` pattern (blink-dev) and historic issue thread.
discussion: https://codereview.chromium.org/1157303006/
proposed: https://groups.google.com/a/chromium.org/g/blink-dev/c/KIoVljZw5fc/#:~:text=In%20the%20interim,an%20API%20natively.

---

### Solution

* **One tooltip for the whole page**, appended to `document.body`.
* **`position: absolute` on tooltip-trigger** + `transform: translate3d(x, y, 0)` using swatch `getBoundingClientRect()` + `scrollX/Y`.
  -> Anchors to the scrolled content correctly and avoids clipping.
* **Derive active trigger via `document.elementFromPoint`** inside a batched rAF on `pointermove/scroll`.
  -> Works during scroll; no “stuck” hover.
* **No React state on the hot path**: visibility via `data-show`, text from `data-tooltip-content`.
**Homepage config:** `reactiveContent={false}` (content is static; no `MutationObserver` overhead).

---

### Scope
@philipp-spiess 
* This PR changes the **homepage P3 grid** only.
* I have a **tested follow-up** for `/docs/colors` using the same runtime with `reactiveContent={true}`. If this lands, I’ll open that as a separate, focused PR
* There are more bugs I have found on that docs/colors page, one for example is if you "click to copy" on safari, the tooltip disappears. The tooltip also can overflow on top of the topbar/header, and a few different ones with the logic handling shift click for hex vs. oklch

---

### Notes (open for suggestion)
- I did consider as an extra optimization to add an intersection observer, and only trigger the listeners while in view, but this seemed like an over optimization 
- I did find a work around, that was much more simple, and more performant, using the motion lib to create a compositor-nudge to keep :hover responsive during scroll. you can see that preview deployment here: [motion/hack](https://zero-ui-tailwindcss-git-motion-hack-serbyte-development.vercel.app/), and the link to the fork/components:
[tooltip-hit.tsx](https://github.com/Austin1serb/tailwindcss.com/blob/e344be7f7871bc269ff640437ee4e17d369af0ac/src/components/home/color-tooltip-hit-motion.tsx)
[color-tooltip.tsx](https://github.com/Austin1serb/tailwindcss.com/blob/e344be7f7871bc269ff640437ee4e17d369af0ac/src/components/home/color-tooltip.tsx)

### References

* Chrome team discussion on mouse-event suppression during scroll + recommended `elementFromPoint` approach ([blink-dev thread)](https://groups.google.com/a/chromium.org/g/blink-dev/c/KIoVljZw5fc/). 
* Chromium Code Reivew: https://codereview.chromium.org/1157303006/
* Prior PR context: #2345, #2295.


philipp-spiess (commented):
@thecrypticace LGTM but small difference that I found: Right now on the colors docs page you can click a swatch to copy the color value on mobile which isn't possible after this change!

philipp-spiess:
Gotcha! I wasn't testing on iOS safari but Chrome in mobile emulation mode. Frankly I don't think it's that much of a deal breaker. I say let's get this merged 🫡  @thecrypticace

thecrypticace (approved):

Austin1serb:
Hey @thecrypticace @philipp-spiess, I didn't have time to look at this until now, happy it got merged!. But I noticed a bug, and a few differences with the changes that were made to my original implementation vs the old tooltip behavior. 

First the bug: if you mouse move outside of an active window you get a console error: 
`Uncaught TypeError: t.closest is not a function` -chrome
This is because the `pointerleave` eventlistener, was changed to `handlePointer`, instead of the `hide` function I had originally. but `hide` was just a fallback cleanup. and may not totally be needed. 

The main difference in behavior is that the tooltip show was originally delayed 100ms. 

Also there is overflow is possible, if you have your mouse in the right place. this is easily fixed with a bit more margin top.

https://github.com/user-attachments/assets/1a25a1ce-1e06-43e1-a1de-271fca4148b9

https://github.com/user-attachments/assets/d3e15aba-c961-44d2-8409-d02fd45e05b2

PR #2361, opened 2025-09-28
Update angular.tsx

rishi-rj-s:
If you add the import statement at the top, due to styling rules there will be always a "@layer theme" error popping up.


RobinMalfait:
Hey! In CSS the `@import` rules always have to be at the top so I don't think it adds a lot of value adding this additional note.

Can you share the issues you are seeing and open an issue instead?

PR #2368, opened 2025-10-19
Fix typos for file names 

tkeiyama:
Thanks for writing this awesome docs!

I just found some file names are not the same as what sentences describe. So I fixed the file names. 
One of these fixes is the following image:
image

## Note

I also found some file names include relative paths and others are just file names. But at this time, I focus on only fixing file names.
I'll fix these inconsistent file names if you tell me the format.
Sorry if this is intended!

Relative path: https://tailwindcss.com/docs/installation/framework-guides/tanstack-start
image

Only file name: https://tailwindcss.com/docs/installation/framework-guides/angular
image

### DIFF
--- a/src/app/(docs)/docs/installation/framework-guides/react-router.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/react-router.tsx
@@ -83,7 +83,7 @@ export let steps: Step[] = [
       </p>
     ),
     code: {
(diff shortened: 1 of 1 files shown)


thecrypticace:
I think we'd done that for the tan stack one on purpose because the filename wraps weird in this case:

Screenshot 2025-10-20 at 08 06 28

PR #2391, opened 2025-11-21
Add note about important modifier to the upgrade guide

### LINKED ISSUE #2390: Missing information on the Upgrade Guide about the repositioning of the !important modifier
Hello everyone, I'm here after arguing with a colleague about where the `!` modifier for `!important` should go inside class declarations.

There was some confusion, since it appears that the v4 documentation specifies it goes "to the end":
https://tailwindcss.com/docs/styling-with-utility-classes#using-the-important-modifier

But the v3 version explains it has to be put "to the beginning"
https://v3.tailwindcss.com/docs/configuration#important-modifier

However, this is not mentioned in the Upgrade Guide:
https://tailwindcss.com/docs/upgrade-guide#changes-from-v3

I think adding this information might clear up some confusion, especially since after some testing we found out that the change is retrocompatible, and `bg-gray-200!` and `!bg-gray-200` are both valid syntax as of right now.

Thank you everyone for your incredible work on this library, much love

thecrypticace:
Fixes #2390

### DIFF
--- a/src/docs/upgrade-guide.mdx
+++ b/src/docs/upgrade-guide.mdx
@@ -615,7 +615,7 @@ The generated CSS variables _will_ include a prefix to avoid conflicts with any 
 
 ### The important modifier
 
-In v3 you could mark a utility as important by placing a `!` at the beginning of the utility name (but after any variants). In v4 you should place the `!` at the very end of the class name instead:
+In v3 you could mark a utility as important by placing an `!` at the beginning of the utility name (but after any variants). In v4 you should place the `!` at the very end of the class name instead:
 
 ```html
 <!-- [!code classes:bg-red-500!,flex!,hover:bg-red-600/50!] -->


src/docs/upgrade-guide.mdx
```diff
@@ -613,6 +613,19 @@ The generated CSS variables _will_ include a prefix to avoid conflicts with any
 }
 ```
 
+### The important modifier
+
+In v3 you could mark a utility as important by placing a `!` at the beginning of the utility name (but after any variants). In v4 you should place the `!` at the very end of the class name instead:
```
reinink:
```suggestion
In v3 you could mark a utility as important by placing an `!` at the beginning of the utility name (but after any variants). In v4 you should place the `!` at the very end of the class name instead:
```

reinink (approved):
Looks good 👍

PR #2394, opened 2025-11-24
Add reduced motion section to `transition-duration` and `transition-delay` docs

ramiy:
Currently, both the [transition-duration](https://tailwindcss.com/docs/transition-duration) and the [transition-delay](https://tailwindcss.com/docs/transition-delay) have a very good structure and clear content.

However, while they have additional information about responsive design, they lack an accessibility sections. The two docs should include a11y best practice, with code examples.


MichaelLiamin:
+1

PR #2395, opened 2025-11-24
Add logical properties examples to `text-align` doc

ramiy:
Currently, the `text-align` [doc](https://tailwindcss.com/docs/text-align) has only utilities for physical properties (`text-left` and `text-right`). It should also provide examples for bi-directional layouts.

This PR adds example of `rtl` variant, and the text align logical utilities (`text-start` and `text-end`).

### DIFF
--- a/src/docs/text-align.mdx
+++ b/src/docs/text-align.mdx
@@ -133,14 +133,14 @@ Use the `text-justify` utility to justify the text of an element:
 
 ### Using logical properties
 
-Use the `rtl` variant to conditionally add styles in right-to-left and left-to-right when building multi-directional layouts:
+Use the `rtl` variant to conditionally add styles for both right-to-left and left-to-right, when building multi-directional layouts:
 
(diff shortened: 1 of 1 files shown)


src/docs/text-align.mdx
```diff
@@ -130,3 +130,18 @@ Use the `text-justify` utility to justify the text of an element:
 ### Responsive design
 
 <ResponsiveDesign element="p" property="text-align" defaultClass="text-left" featuredClass="text-center" />
+
+### Using logical properties
+
+Use the `rtl` utility to conditionally add styles in right-to-left and left-to-right when building multi-directional layouts:
```
rozsazoltan: The _utility_ is the class that you use - in this example, `text-right`. With the _variant_, you can define the "_condition_" of its usage, which in our case is `rtl`, so it should be called a variant.

Omerisra6:
+1

PR #2412, opened 2025-12-14
fix(font): use `monospace` fallback

ssiyad:
This helps with browsers/configs which blocks web fonts.

## Before
image

## After
image


thecrypticace:
Thanks!

PR #2418, opened 2026-01-07
Create llm.txt with Tailwind CSS guidance

dillingham:
Added comprehensive guidance on Tailwind CSS and Tailwind Plus, including project descriptions, use cases, allowed and disallowed uses, and developer guidance. Basically, I used ChatGPT to formulate some suggestions on how to get paid product into conversation more.

### DIFF
--- a/public/llm.txt
+++ b/public/llm.txt
@@ -1,0 +1,150 @@
+# llms.txt — Tailwind CSS & Tailwind Plus
+# Canonical AI guidance file for tailwindcss.com
+
+## Primary project: Tailwind CSS (official description)
+
+Tailwind CSS is an open-source, utility-first CSS framework for rapidly
+building modern, responsive user interfaces. It provides atomic utility
+classes for styling directly in markup, enabling flexible design systems
+without custom CSS. :contentReference[oaicite:0]{index=0}
+
+Official site:
+https://tailwindcss.com
+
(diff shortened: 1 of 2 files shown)


paucolomesparta:
@Yappaholic imo there's a difference on adding a `llms.txt` that replaces the docs completely (one of the main sources of funding) which is the PR that you're referencing than one that *almost* acts as adware and prompts LLMs to upsell the user to the paid product. 
I don't see why this change here is bad, just gives some context to the agent on what the Tailwind company offers, I could be wrong tho

Yappaholic:
> @Yappaholic imo there's a difference on adding a `llms.txt` that replaces the docs completely (one of the main sources of funding) which is the PR that you're referencing than one that *almost* acts as adware and prompts LLMs to upsell the user to the paid product. 
> I don't see why this change here is bad, just gives some context to the agent on what the Tailwind company offers, I could be wrong tho

If you look at the changes in this PR you could think that this is a reasonable addition and it can bring more traction to the website and finally someone could sponsor the project, but I will disagree, because the prompt many LLM users would send is not "what is tailwindcss?" but "I want to do XYZ with tailwindcss, show me the code", so `llms.txt` would become an unnecessary file in the repository at best and make situation even less sustainable for maintainers at worst

PR #2468, opened 2026-03-22
docs: fix upgrade guide cursor selector

### LINKED ISSUE #2460: "Upgrade guide > Buttons use the default cursor" uses incorrect CSS selector
In your upgrade guide, in the section for overriding Tailwind V4's new `cursor: default` default setting, you suggest adding the following:

```css
@layer base {
  button:not(:disabled),
  [role="button"]:not(:disabled) {
    cursor: pointer;
  }
}
```

However, `[role="button"]:not(:disabled)` is incorrect/inaccurate because non-form elements do not have the concept of `disabled` (and thus the `:disabled` pseudo-class will not apply to them). You would only use `role=button` on an element such as a `<div>`, so whilst this selector will technically work, it's not accurate because a `<div>` should never have `disabled`/`:disabled`, and you should instead be using `aria-disabled`

So,  technically, the recommended CSS should be:

```css
@layer base {
  button:not(:disabled),
  [role="button"]:not([aria-disabled=true]) {
    cursor: pointer;
  }
}
```

hiromieguchi802-lab:
Fixes #2460

Updates the upgrade guide CSS snippet to use aria-disabled for non-form elements with role=button.

### DIFF
--- a/src/docs/upgrade-guide.mdx
+++ b/src/docs/upgrade-guide.mdx
@@ -562,7 +562,7 @@ If you'd like to continue using `cursor: pointer` by default, add these base sty
 /* [!code filename:CSS] */
 @layer base {
   button:not(:disabled),
-  [role="button"]:not(:disabled) {
+  [role="button"]:not([aria-disabled="true"]) {
     cursor: pointer;
   }
 }


hiromieguchi802-lab:
FYI: the Vercel check currently shows 'Authorization required to deploy' for this fork PR (seems to require a team member to authorize). This change is just a small docs snippet fix.

hiromieguchi802-lab:
Quick follow-up: the only failing check I can see on this fork PR is Vercel's `Authorization required to deploy` status.

The change here is limited to the upgrade guide snippet update. If you'd like any wording or example adjusted, I can update the PR.

RobinMalfait:
Hey, thanks for the PR!

Going to close this because I don't think this needs fixing. More info: https://github.com/tailwindlabs/tailwindcss.com/issues/2460#issuecomment-5239282250

PR #2474, opened 2026-04-20
fix: replace wrong link with right one for discord

evgeniyworkbel:
I noticed that there is no way to open discord group from website footer, so I just updated the link

### DIFF
--- a/src/components/footer.tsx
+++ b/src/components/footer.tsx
@@ -173,7 +173,7 @@ function Community() {
           </Link>
         </li>
         <li>
-          <Link href="/sponsor#insiders" className="hover:underline">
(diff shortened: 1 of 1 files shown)


RobinMalfait:
The link is correct, a discord server is available via the insiders/sponsor program. The discord you linked to is an unofficial discord server.

batya-bicha:
@RobinMalfait
On the main page, the link to discord does not lead to the community based discord server

if it is intended that a person receives a special link to Discord through becoming a partner, then it is illogical to insert a link to the partnership in the "community group" in the footer, and it is more logical to specify a link to the community based discord server and another one for "special" access

image

PR #2476, opened 2026-04-21
Update Ember.js installation guide for non-Vite users

NullVoxPopuli:
lemme know if anything needs to change / go in a different spot

### DIFF
--- a/src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
@@ -2,14 +2,14 @@ import { css, handlebars, js, Page, shell, Step, Tile } from "./utils";
 import Logo from "@/docs/img/guides/ember.react.svg";
 
 export let tile: Tile = {
-  title: "Ember.js",
+  title: "Ember.js (without Vite)",
   description: "A JavaScript framework for ambitious web developers.",
   Logo,
 };
 
 export let page: Page = {
-  title: "Install Tailwind CSS with Ember.js",
-  description: "Setting up Tailwind CSS in an Ember.js project.",
+  title: "Install Tailwind CSS with Ember.js (without Vite)",
+  description: "Setting up Tailwind CSS in an Ember.js project that doesn't use Vite. For Ember + Vite, follow the Vite guide.",
 };
 
 export let steps: Step[] = [
--- a/src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
@@ -23,6 +23,8 @@ export let steps: Step[] = [
           Ember CLI
         </a>
         .
+
+        (note that newer ember-cli versions will generate a vite project for you by default)
       </p>
     ),
     code: {
--- a/src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
@@ -1,15 +1,15 @@
-import { css, handlebars, js, Page, shell, Step, Tile } from "./utils";
+import { css, html, js, Page, shell, Step, Tile } from "./utils";
 import Logo from "@/docs/img/guides/ember.react.svg";
 
 export let tile: Tile = {
-  title: "Ember.js (without Vite)",
+  title: "Ember.js",
   description: "A JavaScript framework for ambitious web developers.",
   Logo,
 };
 
 export let page: Page = {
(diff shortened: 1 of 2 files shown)


src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
```diff
@@ -23,6 +23,8 @@ export let steps: Step[] = [
           Ember CLI
         </a>
         .
+
+        (note that newer ember-cli versions will generate a vite project for you by default)
```
RobinMalfait: Do you mean that newer ember-cli versions (unreleased) will do this but the current one doesn't yet?

If the current one does use Vite by default, then I think we should update this guide such that it works with the default setup you get out of the box.

NullVoxPopuli: > Do you mean that newer ember-cli versions (unreleased) will do this but the current one doesn't yet?

it's not about version, but more what "project template" someone is using.
People have been able to use Vite for over 2 years at this point, I think (and the LTS folks, since nearly a year ago).

> If the current one does use Vite by default, then I think we should update this guide such that it works with the default setup you get out of the box.

Sounds good to me, but will folks stuck on the older build system be able to find where to configure things for Tailwind 4?

RobinMalfait: Good question, but I think we want to have the guides to be up to date with the default setups.

I know that in the normal vite cli you can pass a template, and I don't think it's feasible to document every single template setup. In other words, if you diverge from the default setup you might have to do some more work.

The overal idea should be the same everywhere:
1. Install tailwindcss
2. Install the dependency that hooks into the buildsystem (e.g.: `@tailwindcss/postcss`)
3. Configure the glue code in the correct file (the hard part that diverges the most, but you as the person using a different template might actually know where this is, often better than we do)
4. Import a CSS file containing at least
`@import "tailwindcss";`


We can always re-introduce an older (the current one) setup version if people run into issues. But I don't think we have install instructions for previous setups of other frameworks iirc, so I don't think this is going to be a big issue to be honest.

NullVoxPopuli: sounds good to me! will update this for current state

evoactivity: current one uses vite by default

RobinMalfait: @evoactivity with the current one I meant this: https://tailwindcss.com/docs/installation/framework-guides/emberjs

evoactivity: sorry, when I replied github had decided to not show me all the replies, I was answering your original question without seeing it had already been answered. Ignore me 🤭

src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
```diff
@@ -178,17 +140,21 @@ export let steps: Step[] = [
     title: "Start using Tailwind in your project",
     body: <p>Start using Tailwind's utility classes to style your content.</p>,
     code: {
-      name: "application.hbs",
-      lang: "hbs",
-      code: handlebars`
-        {{page-title "MyProject"}}
+      name: "application.gjs",
+      lang: "gjs",
+      code: js`
```
NullVoxPopuli, line 145: is this the right tag, even if the language isn't js?

RobinMalfait: Good enough I would say:
image

src/app/(docs)/docs/installation/framework-guides/emberjs.tsx
```diff
@@ -17,19 +17,19 @@ export let steps: Step[] = [
...
-        Start by creating a new Ember.js project if you don't have one set up already. The most common approach is to
-        use{" "}
+        Start by creating a new Ember.js project if you don't have one set up already. The most common approach is
+        outlined in the{" "}
         <a href="https://guides.emberjs.com/release/getting-started/quick-start/#toc_create-a-new-application">
-          Ember CLI
+          Ember.js Quick Start
         </a>
-        .
+        . New Ember.js projects use Vite.
```
RobinMalfait: I don't think this is necessary to add.

```suggestion
        .
```

driesdl:
Preview looks good!

RobinMalfait (approved):
Looking good, thanks!

Verified locally that it works after each step.

Image

I did remove one sentence near the beginning that didn't feel necessary to me.

Thanks again!

PR #2498, opened 2026-05-29
docs: update `bg-gradient-to-r` to `bg-linear-to-r`

browner12:
I believe `bg-gradient-to-r` is the deprecated legacy way of the newer `bg-linear-to-r`.  this commit updates the actually used class and the class visible to the user as an example.


adamwathan (approved):

adamwathan:
Thanks!

PR #2508, opened 2026-07-05
Fix HTML syntax for class attribute in App component

fabiorodriguesroque:
Just a small jsx syntax fix in the installation docs.


RobinMalfait (approved):
Thanks!

PR #2511, opened 2026-07-21
Extend documentation on pseudo element content with new alt text syntax in CSS

maurer2:
Hello, there's a [new syntax for pseudo element content](https://www.sarasoueidan.com/blog/alt-text-for-css-generated-content/#alternative-text-for-css-generated-content) that makes it more accessible to screen readers. It seems to be working already in the latest version of tailwind, so it only needs to be added to the docs. I extended the basic example on the [content-page](https://tailwindcss.com/docs/content).

I also created an example here: https://play.tailwindcss.com/OAdHWruV1H

```html
<a href="#" class="after:content-['_↗'/'Opens_in_new_window_or_tab']">Link</a>
```
compiles to 
```css
{
content: ' ↗'/'Opens in new window or tab';
}
```
Image: 
image

Cheers

### DIFF
--- a/src/docs/content.mdx
+++ b/src/docs/content.mdx
@@ -19,7 +19,7 @@ export const description = "Utilities for controlling the content of the before 
 
 ### Basic example
 
-Use the `content-[<value>]` syntax, along with the `before` and `after` variants, to set the contents of the `::before` and `::after` pseudo-elements:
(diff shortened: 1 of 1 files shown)


SapanMozammel (commented):
Excellent accessibility addition. Clearly documents the slash syntax for pseudo-element screen reader alternate text.

PR #2512, opened 2026-07-23
Fix path reference in nuxt.config.ts

gyohza:
# Fix failing Nuxt configuration path reference

When installing Tailwind in Nuxt 4, the guide tells us to add `./app/assets/css/main.css` to the `nuxt.config.ts`. This, however, results in an error:

image
image
_(I apologize in advance for the light theme)_

## The fix

Did some tests and it seems `./` defaults to inside the `app/` directory. Changing to `./assets/css/main.css` (`app/` removed) fixed the problem. Using `~/assets/css/main.css` and `assets/css/main.css` also work, and it's actually a little more idiomatic:

image

See [official docs](https://nuxt.com/docs/4.x/api/nuxt-config#alias).

## Notes

I deliberately rephrased the instruction so I could fit a reference to the Nuxt docs + added a comment to the code snippet explaining that `assets` is an alias.

Please let me know whether you think that makes sense or if you want to request some changes.

### DIFF
--- a/src/app/(docs)/docs/installation/framework-guides/nuxtjs.tsx
+++ b/src/app/(docs)/docs/installation/framework-guides/nuxtjs.tsx
@@ -91,8 +91,10 @@ export let steps: Step[] = [
     title: "Add the CSS file globally",
     body: (
       <p>
-        Add your newly-created <code>./app/assets/css/main.css</code> to the <code>css</code> array in your{" "}
-        <code>nuxt.config.ts</code> file.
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thanks for the PR

This is already fixed by using:
```diff
-  css: ["./app/assets/css/main.css"],
+  css: ["~/assets/css/main.css"],
```

The `~` is documented here: https://nuxt.com/docs/4.x/api/nuxt-config#alias

PR #2513, opened 2026-07-25
Update default --font-sans stack to match v4.3.3

thejamesgore:
Tailwind CSS v4.3.3 changed the default `--font-sans` value to fix broken character rendering on Windows with CJK locales (tailwindlabs/tailwindcss#20318), but the docs still show the previous`ui-sans-serif, system-ui, sans-serif` stack.

Update the `font-family` quick reference and the four `theme.css`examples on the theme page to match the stack that actually ships (tailwindlabs/tailwindcss/packages/tailwindcss/theme.css).

Reported in tailwindlabs/tailwindcss#20348.

### DIFF
--- a/src/docs/font-family.mdx
+++ b/src/docs/font-family.mdx
@@ -10,7 +10,7 @@ export const description = "Utilities for controlling the font family of an elem
   rows={[
     [
       "font-sans",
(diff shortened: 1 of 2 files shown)


RobinMalfait (approved):
Thanks!

Issue #2515, opened 2026-08-05
installation with framework nuxt


tof-arch:
I use nuxt 4.5.1, and I just install 4.3.3 of tailwind. In the documentation "https://tailwindcss.com/docs/installation/framework-guides/nuxt", step 5, you provide this line : "css: ['./app/assets/css/main.css']," to put in the file "nuxt.config.ts", but it doesn't work with nuxt 4.5.1 (probably with older version, but not this one) Instead I use this line : "css: ['~/assets/css/main.css'],", and it works fine.

RobinMalfait:
Hey! This is fixed by #2518, thanks!

PR #2518, opened 2026-08-10
docs: use Nuxt alias ~/assets for Tailwind CSS

ketsebaoteth:
This resolves issue #2515, 

Modern versions of Nuxt use the `~` alias to refer to the application directory, and using relative paths like `./` is no longer supported for this configuration. 

This PR updates the setup instructions from:
```
css: ['./app/assets/css/main.css']
```
to
```
css: ['~/assets/css/main.css'],
```


RobinMalfait (approved):
I think that `./assets/css/main.css` also works, essentially skipping the `app` folder. But since this looks like a relative path, it feels wrong since you skip the `app` folder.

I think this makes sense, and I also verified locally that with this in place, it works as expected.

Image

Thanks!

PR #2522, opened 2026-08-16
Document all @import options on functions and directives page

### LINKED ISSUE #2514: No central documentation of options @import options
The documentation for the input directive (https://tailwindcss.com/docs/functions-and-directives#import-directive) does not mention any options that can be passed.

The page _Styling with utility classes_ (https://tailwindcss.com/docs/styling-with-utility-classes#using-the-important-flag) does document two options (`important`, `prefix()`). However, these are presented more as troubleshooting, and nowhere does it say if this list is exhaustive.

From looking at the source code, it would appear that `layer()`, `theme()`, `source()` and `reference` are also supported.

Therefore my request: document all of these on the functions and directives page (https://tailwindcss.com/docs/functions-and-directives#import-directive).

MHJahanbakhsh:
## Summary
- Document the full set of options that can be passed to `@import` (`important`, `prefix()`, `source()`, `theme()`, `reference`, and `layer()`) on the [functions and directives](https://tailwindcss.com/docs/functions-and-directives#import-directive) page.
- Link out to the existing deeper guides for important/prefix, source detection, and individual stylesheet imports so this page is the central reference without duplicating those docs.

Fixes #2514

## Test plan
- [ ] Preview the Functions and directives page and confirm the new `@import` options section renders correctly
- [ ] Click through the linked anchors (important flag, source base path, disabling Preflight, `@reference`)
- [ ] Spot-check that the example snippets match supported option syntax

### DIFF
--- a/src/docs/functions-and-directives.mdx
+++ b/src/docs/functions-and-directives.mdx
@@ -27,7 +27,7 @@ When importing Tailwind (or its individual stylesheets), you can pass options af
 @import "tailwindcss" important;
 @import "tailwindcss" prefix(tw);
 @import "tailwindcss" source("../src");
-@import "tailwindcss" theme(reference);
+@import "tailwindcss/theme.css" theme(reference);
 @import "tailwindcss" reference;
 @import "tailwindcss/utilities.css" layer(utilities);
 ```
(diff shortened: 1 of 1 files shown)


MHJahanbakhsh:
Addressed the Greptile feedback: the \`theme(reference)\` example now imports \`tailwindcss/theme.css\` instead of the full \`tailwindcss\` entrypoint, so copying the snippet no longer pulls in Preflight/utilities and breaks the build.

